Issue #18, opened 2021-03-16
TLS termination hangs in latest commit


regbo:
I am not sure why, but the latest commit of this plugin hangs when terminating TLS connections. When using the below config, this build will hang indefinitely:

 xcaddy build --with github.com/mholt/caddy-l4

This build however, will not:

xcaddy build --with github.com/mholt/caddy-l4@6587f40d4eb632d15f295cf9791d7d145434d0da

No errors are reported in logs. For what it's worth I'm running this on Windows/

```json
{
	"logging": {
		"logs": {
			"default": {
				"level": "DEBUG"
			}
		}
	},
	"apps": {
		"tls": {
			"certificates": {
				"automate": ["mongodb.127.0.0.1.nip.io", "tendis.127.0.0.1.nip.io", "webserver.127.0.0.1.nip.io"]
			},
			"automation": {
				"policies": [{
						"issuer": {
							"module": "internal"
						},
						"key_type": "rsa4096"
					}
				]
			}
		},
		"layer4": {
			"servers": {
				"server_0": {
					"listen": ["0.0.0.0:443"],
					"routes": [{
							"match": [{
									"tls": {
										"sni": ["mongodb.127.0.0.1.nip.io"]
									}
								}
							],
							"handle": [{
									"handler": "tls",
									"connection_policies": []
								}, {
									"handler": "proxy",
									"upstreams": [{
											"dial": ["localhost:27017"]
										}
									]
								}
							]
						}, {
							"match": [{
									"tls": {
										"sni": ["tendis.127.0.0.1.nip.io"]
									}
								}
							],
							"handle": [{
									"handler": "tls",
									"connection_policies": []
								}, {
									"handler": "proxy",
									"upstreams": [{
											"dial": ["localhost:51002"]
										}
									]
								}
							]
						}, {
							"match": [{
									"tls": {
										"sni": ["webserver.127.0.0.1.nip.io"]
									}
								}
							],
							"handle": [{
									"handler": "tls",
									"connection_policies": [{
											"alpn": ["http/1.1"]
										}
									]
								}, {
									"handler": "proxy",
									"upstreams": [{
											"dial": ["localhost:8282"]
										}
									]
								}
							]
						}
					]
				}
			}
		}
	}
}

```

mholt:
What is a command (preferably `curl -v`) or some code that replicates the hang?

hslatman:
We've come across this issue too on Ubuntu. Don't have a reproducer available (yet).

[...]

PR #19, opened 2021-04-15
add downstream PROXY protocol support

contrun:
This PR add downstream PROXY protocol support. Reading from the source of `github.com/mastercactapus/proxyprotocol`, it seems to me that the matcher will not blockingly wait even if there is not enough data. But I need to verify that and find a way to reuse `github.com/mastercactapus/proxyprotocol`.

### DIFF
--- a/modules/l4proxyprotocol/matcher.go
+++ b/modules/l4proxyprotocol/matcher.go
@@ -1,0 +1,105 @@
+// Copyright 2021 contrun
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package l4proxyprotocol
+
+import (
+	"errors"
+	"net"
+	"time"
+
+	"github.com/caddyserver/caddy/v2"
+	"github.com/mastercactapus/proxyprotocol"
+	"github.com/mholt/caddy-l4/layer4"
+	"go.uber.org/zap"
+)
+
+func init() {
+	caddy.RegisterModule(MatchPROXYProtocol{})
+}
+
+// MatchPROXYProtocol is able to match PROXY protocol connections.
+type MatchPROXYProtocol struct {
+	// Note by default, we allow all client ips.
+	AllowedRanges []string `json:"allowed_ranges,omitempty"`
+
+	allowedSubnets []*net.IPNet
+	logger         *zap.Logger
+}
+
+// CaddyModule returns the Caddy module information.
+func (MatchPROXYProtocol) CaddyModule() caddy.ModuleInfo {
+	return caddy.ModuleInfo{
+		ID:  "layer4.matchers.proxyprotocol",
+		New: func() caddy.Module { return new(MatchPROXYProtocol) },
+	}
+}
+
+func (m MatchPROXYProtocol) isIPAllowed(ip net.IP) bool {
+	if len(m.allowedSubnets) == 0 {
+		return true
+	}
+	for _, n := range m.allowedSubnets {
+		if n.Contains(ip) {
+			return true
+		}
+	}
+	return false
+}
+
+// Provision sets up the module.
+func (t *MatchPROXYProtocol) Provision(ctx caddy.Context) error {
+	t.logger = ctx.Logger(t)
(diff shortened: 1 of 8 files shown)


modules/l4proxyprotocol/matcher.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	t.allowedSubnets = cidrs
+	return nil
+}
+
+// Match returns true if the connection is a PROXY protocol handshake.
+// Then saves the terminated proxy protocol to `terminated_proxy_connection`.
+func (m MatchPROXYProtocol) Match(cx *layer4.Connection) (bool, error) {
+	// We use `cx.Conn` instead of `cx` here as proxyprotocol already only peeks into the packet.
+	conn := proxyprotocol.NewConn(cx.Conn, time.Time{})
```
contrun, line 76: Here proxyprotocol will read a few bytes from the underlying OS FD. It is OK if we eventually use `conn`. But other matchers will not work as there will be a few bytes discarded. I can't `conn := proxyprotocol.NewConn(cx, time.Time{})` either. It would cause a infinite calling loop of `conn.Read()` -> `cx.Conn.Read()` -> `conn.Read()` as both `layer4.Connection` and `proxyprotocol.NewConn` embed the same `net.Conn`. Unfortunately `github.com/mastercactapus/proxyprotocol` and other golang PROXY protocol library do not provide a easy way to read from a `io.Reader` and then return the PROXY protocol's header or an error. This will make our life easier as we already have buffered IO with `layer4.Connection`.  I find it hard to reuse existing libraries. What are @mholt and @mastercactapus 's opinions on this issue? Should I just write a parser for PROXY protocol just like what @mholt did for TLS?

mastercactapus: If you can get access to the underlying buffered reader, you could use `Parse`: `func Parse(r *bufio.Reader) (Header, error)` that should work as-is.

Otherwise, I could look at adding a `ReadHeader(io.Reader) (Header, error)` function to the library. The difficulty is in the need to call `Read` byte-by-byte for the `V1` header since it's newline terminated.

In either case, you'd need to layer or at least re-use the same `bufio.Reader` to preserve the extra bytes. Alternatively, a `ReadHeader` method could return any buffered bytes as a slice, similar to `json.Decoder`

mholt: I would recommend passing in `cx` here, since a few bytes may be buffered from a prior matcher, as you mentioned.

Indeed, though, this can cause a bit of a hanging problem as described in #18. I hope to be pushing a fix for that today, so follow that issue and I'll try to follow up here if I remember. :)

mholt: @contrun Ok, fixed. Here's how it'll look in your PR:

- Pass `cx` into `NewConn()`.
- If you need a separate Connection that doesn't wrap itself, do something like `cx = cx.Wrap(conn)`.

Let me know how that goes!

contrun: @mholt Thank you for your work in https://github.com/mholt/caddy-l4/commit/5f9948fe76ae19009258465fd66da1390aa5ecce . I tried the `Wrap` function, it's working great. I added a `json` configuration which works on my machine. The only remaining issue for me is that `conn := proxyprotocol.NewConn(cx, time.Time{})` seems to disable timeout configuration of the underlying connection. I think the signature here https://pkg.go.dev/github.com/pires/go-proxyproto#NewConn is more reasonable than https://pkg.go.dev/github.com/mastercactapus/proxyprotocol?utm_source=godoc#NewConn , as it does not require us to set a timeout interval. I'd like to hear @mastercactapus 's opinion.

mastercactapus: Hmm, if no deadline is set (or is zero) expected behavior IMO would be to not set (thus not un-set) the deadline, I'd consider that a bug.
https://github.com/mastercactapus/proxyprotocol/issues/4

I'll try to get a fix pushed up for that.

mastercactapus: Have this in a branch: https://github.com/mastercactapus/proxyprotocol/tree/use-io-reader

- `Parse` now takes an `io.Reader` and using `ReadByte` if available otherwise falling back to `Read`
- Deadline methods will no longer be called if `time.Time{}` is passed to `NewConn`
- A new `WrapConn` method will simply do the wrap and block/read/parse the header immediately and is intended to replace `NewConn`
- A new `WrapConnReader` can be used to specify the reader for the connection (e.g., use existing `*bufio.Reader`)

Trying to handle the deadline stuff with delayed parsing just seems like a bad idea as I'm looking at it. Especially since the deadline methods should be able to be used concurrently, and thus there are issues with synchronizing that on top of determining expected behavior.

Not 100% on it all yet, would love feedback on what options/changes seem to fit best.

mholt:
@contrun Do you want to finish this up?

mholt:
Closing due to inactivity and merge conflict

Issue #20, opened 2021-04-19
Nested matching and handling


contrun:
It is a common case that the client sends the HTTP traffic in a TLS session and the proxy forwards it upstream in PROXY protocol. I want to match PROXY protocol and then match TLS session, if the match fails then we should fallback to next route. To the best of my knowledge, there is currently no way to achieve such effect. I propose we implement a special handler called `nested` to handle such case. 

Below is an example configuration. Assume that we have already implemented a matcher called `proxyprotocol` which matches all traffic in PROXY protocol and a handler called `proxyprotocol` which turns PROXY protocol traffic into normal TCP traffic. The below configuration instructs `caddly-l4` to first matches proxy protocol, then unwraps PROXY protocol traffice and handles the unwrapped traffic with the specified routes, which in turn try to match it with tls and then terminate tls, if that fails, match it with the http matcher.

If the entire nested matcher failed to match the connection, we should fallback to the next route. My configuration below makes `caddy-l4` accept both traffic in PROXY protocol and in normal TCP connection.
```
{
  "apps": {
    "layer4": {
      "servers": {
        "example": {
          "listen": ["0.0.0.0:5000"],
          "routes": [
            {
              "match": [
                {
                  "proxyprotocol": {}
                }
              ],
              "handle": [
                {
                  "handler": "proxyprotocol"
                },
                {
                  "handler": "nested",
                  "routes": [
                    {
                      "match": [
                        {
                          "tls": {}
                        }
                      ],
                      "handle": [
                        {
                          "handler": "tls"
                        },
                        {
                          "handler": "proxy",
                          "upstreams": [{ "dial": ["localhost:8080"] }]
                        }
                      ]
                    },
                    {
                      "match": [
                        {
                          "http": [{ "host": ["example.com"] }]
                        }
                      ],
                      "handle": [
                        {
                          "handler": "proxy",
                          "upstreams": [{ "dial": ["localhost:8080"] }]
                        }
                      ]
                    }
                  ]
                }
              ]
            },
            {
              "match": [
                {
                  "tls": {}
                }
              ],
              "handle": [
                {
                  "handler": "tls"
                },
                {
                  "handler": "proxy",
                  "upstreams": [{ "dial": ["localhost:8080"] }]
                }
              ]
            },
            {
              "match": [
                {
                  "http": [{ "host": ["example.com"] }]
                }
              ],
              "handle": [
                {
                  "handler": "proxy",
[...]

PR #30, opened 2021-09-26
l4proxy: add proxy_protocol matcher

MadhuvanthG:
adds capability for part of #20 

- not sure yet on how to build caddy locally that includes changes to a module. This instruction in [README](https://github.com/mholt/caddy-l4#compiling) doesn't work me, I don't see the layer4 modules, if I build that way
> In the project folder, run xcaddy just like you would run caddy. For example: xcaddy list-modules --versions (you should see the layer4 modules).

-  Not familiar with how could we test this module which operates at layer4? Do we use certain tools to simulate traffic? Although this matcher is very simple, would like to test that it works, to be sure.

### DIFF
--- a/modules/l4proxy/matcher.go
+++ b/modules/l4proxy/matcher.go
@@ -1,0 +1,49 @@
+// Copyright 2020 Matthew Holt
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package l4proxy
+
+import (
+	"bufio"
+
+	"github.com/caddyserver/caddy/v2"
+	"github.com/mastercactapus/proxyprotocol"
+	"github.com/mholt/caddy-l4/layer4"
+)
+
+func init() {
+	caddy.RegisterModule(MatchPROXY{})
+}
+
+type MatchPROXY struct{}
+
+// CaddyModule returns the Caddy module information.
+func (MatchPROXY) CaddyModule() caddy.ModuleInfo {
+	return caddy.ModuleInfo{
+		ID:  "layer4.matchers.PROXY",
+		New: func() caddy.Module { return new(MatchPROXY) },
+	}
+}
+
+// Match returns true if the connection looks like SSH.
+func (m MatchPROXY) Match(cx *layer4.Connection) (bool, error) {
+	_, err := proxyprotocol.Parse(bufio.NewReader(cx))
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// Interface guard
+var _ layer4.ConnMatcher = (*MatchPROXY)(nil)
--- a/modules/l4proxy/matcher.go
+++ b/modules/l4proxy/matcher.go
@@ -31,21 +31,21 @@ var (
 )
 
 func init() {
-	caddy.RegisterModule(MatchPROXY{})
+	caddy.RegisterModule(MatchProxyProtocol{})
 }
 
-type MatchPROXY struct{}
+type MatchProxyProtocol struct{}
 
 // CaddyModule returns the Caddy module information.
-func (MatchPROXY) CaddyModule() caddy.ModuleInfo {
+func (MatchProxyProtocol) CaddyModule() caddy.ModuleInfo {
 	return caddy.ModuleInfo{
-		ID:  "layer4.matchers.ProxyProtocol",
-		New: func() caddy.Module { return new(MatchPROXY) },
+		ID:  "layer4.matchers.proxy_protocol",
+		New: func() caddy.Module { return new(MatchProxyProtocol) },
 	}
 }
 
 // Match returns true if the connection looks like it is using the Proxy Protocol.
-func (m MatchPROXY) Match(cx *layer4.Connection) (bool, error) {
+func (m MatchProxyProtocol) Match(cx *layer4.Connection) (bool, error) {
 	p := make([]byte, 4)
 	_, err := io.ReadFull(cx, p)
 	if err != nil {
@@ -87,4 +87,4 @@ func (m MatchPROXY) Match(cx *layer4.Connection) (bool, error) {
 }
 
 // Interface guard
-var _ layer4.ConnMatcher = (*MatchPROXY)(nil)
+var _ layer4.ConnMatcher = (*MatchProxyProtocol)(nil)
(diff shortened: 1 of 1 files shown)


modules/l4proxy/matcher.go
```diff
@@ -0,0 +1,49 @@
...
+func init() {
+	caddy.RegisterModule(MatchPROXY{})
+}
+
+type MatchPROXY struct{}
+
+// CaddyModule returns the Caddy module information.
+func (MatchPROXY) CaddyModule() caddy.ModuleInfo {
+	return caddy.ModuleInfo{
+		ID:  "layer4.matchers.PROXY",
```
MadhuvanthG, line 34: modules name in caps (PROXY) helps provide clarity on what this matches against, but this can also be confusing, I imagine. Any suggestions on naming?

modules/l4proxy/matcher.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/caddyserver/caddy/v2"
+	"github.com/mastercactapus/proxyprotocol"
+	"github.com/mholt/caddy-l4/layer4"
+)
+
+func init() {
+	caddy.RegisterModule(MatchPROXY{})
+}
+
+type MatchPROXY struct{}
```
MadhuvanthG, line 29: Do we want the ability to match against the source IP/Port? Right now, it matches a connection, as long as it finds a PROXY protocol header.

modules/l4proxy/matcher.go
mholt: Yeah, good point. Maybe `MatchProxyProtocol` would be more clear and less loud.

modules/l4proxy/matcher.go
mholt: That would be a separate matcher that already exists:

https://github.com/mholt/caddy-l4/blob/3cfcafe708834b071dae82a5f75224e55baef230/layer4/matchers.go#L95-L100

modules/l4proxy/matcher.go
```diff
@@ -0,0 +1,49 @@
...
+
+// CaddyModule returns the Caddy module information.
+func (MatchPROXY) CaddyModule() caddy.ModuleInfo {
+	return caddy.ModuleInfo{
+		ID:  "layer4.matchers.PROXY",
+		New: func() caddy.Module { return new(MatchPROXY) },
+	}
+}
+
+// Match returns true if the connection looks like SSH.
```
mholt, line 39:
```suggestion
// Match returns true if the connection looks like it is using the Proxy Protocol.
```

modules/l4proxy/matcher.go
```diff
@@ -0,0 +1,49 @@
...
+func (m MatchPROXY) Match(cx *layer4.Connection) (bool, error) {
+	_, err := proxyprotocol.Parse(bufio.NewReader(cx))
```
mholt, line 41: As much as I like this library, we don't need to use it for this. You basically just need to read the first few bytes to see if they are one of two values. It's just a few lines of Go code.

Reference:
https://github.com/mastercactapus/proxyprotocol/blob/ae65f2ce73c948fdd0d8f00ffaa6c519373c38bf/parse.go

Basically a combination of that file and what we do for the SSH matcher. But instead of SSH, look for Proxy protocol bytes.

mholt (changes requested):
Thanks for working on this! It can be improved and we can drop a dependency in this file.

modules/l4proxy/matcher.go
MadhuvanthG: I'm going ahead with `ProxyProtocol` since `Match` prefix would be redundant. Let me if you have different thoughts.

modules/l4proxy/matcher.go
MadhuvanthG: Oops sorry, good catch.

modules/l4proxy/matcher.go
MadhuvanthG: Sure, I did see that matcher, but if Caddy is behind another proxy, then this would match against the proxy's IP rather than the original connection's IP which is available in the PROXY header. 

As I write this comment though, it occurs to me that it might not be a valid use-case to expect Caddy (while placed behind another proxy) to match against original connection's IP addresses.

MadhuvanthG: One of the reasons why I used that library initially, since it has the logic to parse the PROXY header fully and retrieve source/destination ip address/port.

modules/l4proxy/matcher.go
MadhuvanthG: yeah sure, I changed the implementation now which drops this dependency.

modules/l4proxy/matcher.go
```diff
@@ -0,0 +1,90 @@
...
+			return false, err
+		}
+
+		// 4 bytes were already read, so i == 1 is the 5th byte
+		if i != 1 {
+			return false, err
+		}
+
+		buf = append(buf, 0x00)
+	}
```
MadhuvanthG, line 80: this whole logic is to handle this part of PROXY protocol spec properly. 
> Note that this block contains a null byte at the 5th position, so it must not
> be handled as a null-terminated string.

But when I test which `telnet`, Caddy still closes the connection, so I'm thinking is this the right way to handle or/and am I not testing it properly. Any suggestions would be helpful.

MadhuvanthG: Btw I tested the v1 header with an `echo` handler and the matcher does the job :)

modules/l4proxy/matcher.go
mholt: Ah, well, let's wait and see if an actual use case appears first.

mholt (approved):
LGTM, thanks for contributing this! And sorry for the delay.

modules/l4proxy/matcher.go
mholt: Oh actually, can we rename this to `MatchProxyProtocol` first?

modules/l4proxy/matcher.go
```diff
@@ -0,0 +1,90 @@
...
+func init() {
+	caddy.RegisterModule(MatchPROXY{})
+}
+
+type MatchPROXY struct{}
+
+// CaddyModule returns the Caddy module information.
+func (MatchPROXY) CaddyModule() caddy.ModuleInfo {
+	return caddy.ModuleInfo{
+		ID:  "layer4.matchers.ProxyProtocol",
```
francislavoie, line 42: I think typically we use `snake_case`

```suggestion
		ID:  "layer4.matchers.proxy_protocol",
```

mholt (changes requested):
Oops, seems the rename got passed up

modules/l4proxy/matcher.go
MadhuvanthG: Thanks for raising that, I've changed to snake case now.

modules/l4proxy/matcher.go
MadhuvanthG: Have renamed it.

MadhuvanthG:
@mholt addressed the suggestion on renaming. Let me know if there are any other comments.

Looking forward to getting this merged.

francislavoie (approved):
LGTM! Thanks for updating the PR 😄

mholt (approved):
LGTM, thank you for the changes and for the contribution!

PR #33, opened 2021-11-05
l4xmpp: Add support for matching XMPP connections, match TLS-ALPN

Rijul-A:
1. Add generic ALPN matcher, which is used to match `xmpp-client` for XMPP. Let me know if this is better suited elsewhere (like the main Caddy repository).
2. Add XMPP matcher, which is similar to SSH matcher except it looks for the word `jabber` in the [first 50 bytes](https://github.com/yrutschle/sslh/blob/0e118a109c0407e47c8e43e37662df0f6c0367f2/probe.c#L175). 

Use case: Replacement for forwarding Snikket traffic over port 443 to relevant service; documented [here](https://github.com/snikket-im/snikket-server/blob/master/docs/advanced/reverse_proxy.md#sslh) using `sslh`

### DIFF
--- a/modules/l4tls/alpn_matcher.go
+++ b/modules/l4tls/alpn_matcher.go
@@ -1,0 +1,56 @@
+// Copyright
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package l4tls
+
+import (
+  "crypto/tls"
+
+	"github.com/caddyserver/caddy/v2"
+	"github.com/caddyserver/caddy/v2/modules/caddytls"
+)
+
+func init() {
+	caddy.RegisterModule(MatchALPN{})
+}
+
+type MatchALPN []string
+
+// CaddyModule returns the Caddy module information.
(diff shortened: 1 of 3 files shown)


mholt:
Hi, thanks for this contribution! I am trying to tag a new release of Caddy but I hope to look at this in the next little while.

modules/l4tls/alpn_matcher.go
```diff
@@ -0,0 +1,56 @@
...
+		ID:  "tls.handshake_match.alpn",
+		New: func() caddy.Module { return new(MatchALPN) },
+	}
+}
+
+func (m MatchALPN) Match(hello *tls.ClientHelloInfo) bool {
+  clientProtocols := hello.SupportedProtos
+  if len(clientProtocols) == 0 {
+    return false
+  }
```
mholt, line 42: Is this needed? We can probably omit this right?

mholt (changes requested):
I like this! The XMPP one is a good fit. I'm trying to decide whether the ALPN matcher is a better fit for Caddy's `caddytls` package directly so that other Caddy sites can use it, or if it should be exclusive to this module. Hmm hmm hmm. Most Caddy apps are using HTTP on top of TLS, since most Caddy users use it for its HTTP server. But then again, there could be value in handling a handshake one way based on ALPN versus another way, since that repo is where the TLS server lives.

Hmm...

I guess we can always start with it here and then move it if we need to (it's harder to go the other way around).

Please use `go fmt` to clean up the code, and ensure the copyright is intact. Thanks!

mholt (approved):
Neat, thank you so much for contributing this!

Rijul-A:
I just observed that the SSH matcher must be before the XMPP matcher for the handling to work correctly (matching appears to work fine in either order).

mholt:
@Rijul-A I wonder if that is due/related to https://github.com/mholt/caddy-l4/pull/72.

PR #41, opened 2022-01-19
l4proxy: Read from all upstream connections before fully closing them

### LINKED ISSUE #40: l4proxy: Upstream connections may be closed before reading from them
## Problem

The close of upstream connections:

https://github.com/mholt/caddy-l4/blob/31d74d38aa3a41c1563f25dfbee24ced3d4afae2/modules/l4proxy/proxy.go#L173-L178

may happen before the reading of replies from the upstream connections:

https://github.com/mholt/caddy-l4/blob/31d74d38aa3a41c1563f25dfbee24ced3d4afae2/modules/l4proxy/proxy.go#L256-L263

As a result, **the replies from upstream servers will never be forwarded to the downstream clients**. 

## Reproducer

Given a slow Echo server:

```go
package main

import (
	"io"
	"log"
	"net"
	"time"
)

type SlowConn struct {
	net.Conn
}

func (c SlowConn) Write(p []byte) (n int, err error) {
	time.Sleep(5 * time.Millisecond) // Write slowly
	n, err = c.Conn.Write(p)
	return
}

func main() {
	ln, err := net.Listen("tcp", ":7070")
	if err != nil {
		log.Panicln(err)
	}
	defer ln.Close()

	for {
		conn, err := ln.Accept()
		if err != nil {
			log.Panicln(err)
		}

		go handle(conn)
	}
}

func handle(conn net.Conn) {
	c := SlowConn{Conn: conn}
	if _, err := io.Copy(c, c); err != nil {
		log.Printf("err: %v\n", err)
	}
	conn.Close()
}
```

Everything is OK:

```bash
$ echo "hi" | nc localhost 7070
hi
```

Run Caddy with the following JSON config:

```json
{                                       
        "apps": {                       
                "layer4": {                     
                        "servers": {                    
                                "srv0": {
                                        "listen": [":8080"],
                                        "routes": [                             
                                                {                       
                                                        "handle": [     
                                                                {
                                                                        "handler": "proxy",
                                                                        "upstreams": [
                                                                                {"dial": ["localhost:7070"]}
                                                                        ]
                                                                }
                                                        ]
                                                }
                                        ]
                                }
                        }
                }
        }
}
```

Nothing will be displayed:

```bash
$ echo "hi" | nc localhost 8080
```

RussellLuo:
This closes #40.

### DIFF
--- a/modules/l4proxy/proxy.go
+++ b/modules/l4proxy/proxy.go
@@ -22,7 +22,7 @@ import (
 	"log"
 	"net"
 	"runtime/debug"
-	"sync/atomic"
+	"sync"
 	"time"
 
 	"github.com/caddyserver/caddy/v2"
@@ -247,14 +247,16 @@ func (h *Handler) proxy(down *layer4.Connection, upConns []net.Conn) {
 		downTee = io.TeeReader(downTee, up)
 	}
 
-	// when we are done and have closed connections, set this
-	// flag to 1 so that we don't report errors unnecessarily
-	var done int32
+	var wg sync.WaitGroup
 
 	for _, up := range upConns {
+		wg.Add(1)
+
 		go func(up net.Conn) {
+			defer wg.Done()
+
 			_, err := io.Copy(down, up)
-			if err != nil && atomic.LoadInt32(&done) == 0 {
+			if err != nil {
 				h.logger.Error("upstream connection",
 					zap.String("local_address", up.LocalAddr().String()),
 					zap.String("remote_address", up.RemoteAddr().String()),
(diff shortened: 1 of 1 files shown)


mholt (approved):
Nice, this is much more elegant. And thank you for diving in and figuring out a way to unblock the write-to side! Thanks for the contribution.

(Would you like push access to the repo? Then you can contribute in a branch instead of a fork if you want to.)

RussellLuo:
> (Would you like push access to the repo? Then you can contribute in a branch instead of a fork if you want to.)

Sure, it's my pleasure! Thanks for your trust.

PR #44, opened 2022-01-27
proxy: Close the downstream after all upstreams are closed

### LINKED ISSUE #38: [Feature] close downstream if upstream is closed
HI:

I'm using Caddy as a layer4 load balancer for Kubelet to Kube-APIServer.  And I found Kubelet could be `hang` in some time.

After days of digging, I think I found the reason:
In the scenario of Kubernetes, the Client uses Watch&List to describe resources changing events.  This means that kube-apiserver returns all resources and pushes events only if resources changed. however, after kube-apiserver restarted, the upstream of Caddy is closed, but the downstream is still connected.  At the same time, the client is still waiting for events to come which is impossible. And the worst thing is the client will never find out this connection should be reset.

Kubernetes fix this issue in 1.20 in https://github.com/kubernetes/client-go/issues/374. but in some case, we can't upgrade to the latest version so quickly. 

Here is my proposal:
Maybe we can add an option to the LoadBalancer struct, such as `DisconnectDownstreamPolicy`. And the value could be `Never`、`OnAnyUpsteamClosed` or `OnAllUpstreamClosed`.  the default value is `Never` which keeps the same behavior as current.

How does this sound to you?  And If you are busy, I'm ok to submit a Pull&Request and wait for you to review it. the poc change is working for me. https://github.com/xuzhenglun/caddy-l4/commit/f013a43f63667218527de3c95dbf862e14466e92

RussellLuo:
This closes #38.

### DIFF
--- a/modules/l4proxy/proxy.go
+++ b/modules/l4proxy/proxy.go
@@ -260,8 +260,7 @@ func (h *Handler) proxy(down *layer4.Connection, upConns []net.Conn) {
 		go func(up net.Conn) {
 			defer wg.Done()
 
-			_, err := io.Copy(down, up)
-			if err != nil {
+			if _, err := io.Copy(down, up); err != nil {
 				h.logger.Error("upstream connection",
 					zap.String("local_address", up.LocalAddr().String()),
 					zap.String("remote_address", up.RemoteAddr().String()),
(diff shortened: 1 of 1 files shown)


modules/l4proxy/proxy.go
```diff
@@ -271,20 +270,35 @@ func (h *Handler) proxy(down *layer4.Connection, upConns []net.Conn) {
 		}(up)
 	}
 
-	// read from downstream until connection is closed;
-	// TODO: this pumps the reader, but writing into discard is a weird way to do it; could be avoided if we used io.Pipe - see _gitignore/oldtee.go.txt
-	io.Copy(ioutil.Discard, downTee)
+	downConnClosedCh := make(chan struct{}, 1)
```
xuzhenglun, line 273: Is it necessary? 

because `wg.Wait()`  blocks until all upstream is closed. maybe just close downstream after `wg.Wait`?

And if downstream close before upstream, `io.Copy` in L263 will return, and `defer` in L261 will be executed.

RussellLuo: The basic idea here is to **gracefully close the downstream if the upstream is half-closed**, which is a common case in HTTP (see [9.6. Tear-down](https://httpwg.org/http-core/draft-ietf-httpbis-messaging-latest.html#persistent.tear-down)).

Suppose a scenario that consists of a downstream D (an HTTP client) and an upstream U (an HTTP server). If U performs a half-close (still can read from D), `io.Copy` in L263 will return, thus will finally make `wg.Wait` in L292 return. If we don't wait for the closure of D by `<-downConnClosedCh`, there's no chance for the last data from D to be read by U.

Half-closed connections are also common on the client side when involving tools such as netcat (#40) and FTP.

xuzhenglun: Thanks for your explanation, Really appreciate it!

mholt (approved):
LGTM, let's give this a try. Thank you again very much for the contribution!

PR #45, opened 2022-02-05
Reformat README json to use 2 space indents

Xerkus:
Json in README code blocks is displayed in browser with 8-wide tabs which makes it hard to read.

This PR reformats them to use 2 space indents instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,22 +72,22 @@ A simple echo server:
 
 ```json
 {
-	"apps": {
-		"layer4": {
-			"servers": {
-				"example": {
-					"listen": ["127.0.0.1:5000"],
-					"routes": [
-						{
-							"handle": [
-								{"handler": "echo"}
-							]
-						}
-					]
-				}
-			}
-		}
-	}
+  "apps": {
+    "layer4": {
+      "servers": {
+        "example": {
+          "listen": ["127.0.0.1:5000"],
+          "routes": [
+            {
+              "handle": [
+                {"handler": "echo"}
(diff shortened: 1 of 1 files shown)


RussellLuo:
FYI, this seems to be the same issue as https://github.com/mholt/caddy-ratelimit/issues/1.

Xerkus:
Many editors support editor config

`.editorconfig` file in the root might be picked up by the editor automatically:

```
[README.md]
indent_style = space
indent_size = 2
```

mholt:
Tab width is a rendering preference. GitHub and other editors let you customize the width, and I'm not interested in changing to spaces because everyone has different opinions on this, but at least tabs let each user configure them.

Xerkus:
It was not about the editor. It is about what you see when you open the repository web page page. First thing you see, pretty much.

Readme screenshot

mholt:
If you've changed your tab width settings on GitHub to 2 and it's still rendering them as 8-wide, then that's probably a bug in GitHub...

Xerkus:
Checked css, github does set tab-size css for code files but not for `<pre>` tag in rendered html from markdown.
Does the setting work for you there? I guess I can try to send bug report to github.

francislavoie:
As I've said in the other thread https://github.com/mholt/caddy-ratelimit/issues/1#issuecomment-810534882, I use the https://github.com/sindresorhus/refined-github browser plugin which does fix this. But yeah, it would make sense to ask Github to fix this.

PR #43, opened 2022-01-24
proxyprotocol: Add support for accepting the PROXY protocol

RussellLuo:
This closes #27.

With the following JSON config:

```json
{
	"apps": {
		"layer4": {
			"servers": {
				"srv0": {
					"listen": [":7070"],
					"routes": [
						{
							"handle": [
								{
									"handler": "proxy",
									"proxy_protocol": "v1",
									"upstreams": [
										{"dial": ["localhost:8080"]}
									]
								}
							]
						}
					]
				},
				"srv1": {
					"listen": [":8080"],
					"routes": [
						{
							"handle": [
								{
									"handler": "proxy_protocol"
								},
								{
									"handler": "echo"
								}
							]
						}
					]
				}
			}
		}
	},
	"logging": {
		"logs": {
			"default": {
				"level": "DEBUG"
			}
		}
	}
}
```

I got the result as below:

```bash
$ echo -n "hi" | nc -s 192.168.2.147 localhost 7070                                              
hi
```

and here are the DEBUG logs:

```
2022/01/24 14:43:40.617	DEBUG	layer4.handlers.proxy	dial upstream	{"remote": "192.168.2.147:63372", "upstream": "localhost:8080"}
2022/01/24 14:43:40.617	DEBUG	layer4.handlers.proxy_protocol	received the PROXY protocol header	{"remote": "192.168.2.147:63372", "local": "127.0.0.1:7070"}
2022/01/24 14:43:40.617	DEBUG	layer4	connection stats	{"remote": "127.0.0.1:63373", "read": 49, "written": 2, "duration": 0.000208211}
2022/01/24 14:43:40.617	DEBUG	layer4	connection stats	{"remote": "192.168.2.147:63372", "read": 2, "written": 2, "duration": 0.001086841}
```

---
**UPDATE**: Add an option `allow` as below:

```json
								{
									"handler": "proxy_protocol",
									"allow": ["192.168.0.0/16"]
								},
```

then the PROXY header will not be read by the `proxy_protocol`handler in "srv1":

```bash
$ echo -n "hi" | nc -s 192.168.2.147 localhost 7070
PROXY TCP4 192.168.2.147 127.0.0.1 54566 7070
hi
```

and these are the corresponding logs:

```
2022/01/25 04:36:45.740	DEBUG	layer4.handlers.proxy	dial upstream	{"remote": "192.168.2.147:54566", "upstream": "localhost:8080"}
[...]

PR #52, opened 2022-03-31
subroute: Add support for subrouting

### LINKED ISSUE #20: Nested matching and handling

RussellLuo:
This closes #20.

With the following JSON config:

```json
{
	"apps": {
		"layer4": {
			"servers": {
				"srv0": {
					"listen": ["127.0.0.1:5000"],
					"routes": [
						{
							"match": [
								{
									"http": [
										{"host": ["example.com"]}
									]
								}
							],
							"handle": [
								{
									"handler": "subroute",
									"routes": [
										{
											"match": [
												{
													"http": [
														{"path": ["/1"]}
													]
												}
											],
											"handle": [
												{
													"handler": "proxy",
													"upstreams": [
														{"dial": ["localhost:5001"]}
													]
												}
											]
										},
										{
											"match": [
												{
													"http": [
														{"path": ["/2"]}
													]
												}
											],
											"handle": [
												{
													"handler": "proxy",
													"upstreams": [
														{"dial": ["localhost:5002"]}
													]
												}
											]
										}
									]
								}
							]
						}
					]
				}
			}
		},
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":5001"
					],
					"routes": [
						{
							"handle": [
								{
									"body": "This is 5001",
									"handler": "static_response"
								}
							]
						}
					]
				},
				"srv1": {
					"listen": [
						":5002"
					],
					"routes": [
						{
							"handle": [
								{
									"body": "This is 5002",
									"handler": "static_response"
								}
							]
						}
					]
				}
			}
		}
	}
}
```

We can get the results as below:

```bash
$ curl -H 'Host: example.com' http://127.0.0.1:5000/1
This is 5001
$ curl -H 'Host: example.com' http://127.0.0.1:5000/2
This is 5002
$ curl http://127.0.0.1:5000/1
curl: (52) Empty reply from server
$ curl http://127.0.0.1:5000/2
curl: (52) Empty reply from server
```


imports.go
```diff
@@ -20,7 +20,9 @@ import (
 	_ "github.com/mholt/caddy-l4/modules/l4echo"
 	_ "github.com/mholt/caddy-l4/modules/l4http"
 	_ "github.com/mholt/caddy-l4/modules/l4proxy"
+	_ "github.com/mholt/caddy-l4/modules/l4proxyprotocol"
```
francislavoie, line 23: Oh, was this just not working cause this was missing? Huh.

francislavoie (approved):
LGTM

imports.go
RussellLuo: Yeah, definitely! I just have no idea why it was OK when I tested it in #43 🤔

francislavoie: Well, oops 😅

RussellLuo: Oh, I got it! Because it has been imported by l4proxy:

https://github.com/mholt/caddy-l4/blob/d1f4337132468dfe2d081baba3769dfd1f46fbf8/modules/l4proxy/proxy.go#L31

mholt (approved):
Awesome, thanks for contributing this feature!

PR #54, opened 2022-04-04
server: Don't close underlying UDP conn after handling

### LINKED ISSUE #53: PacketConn (udp) listener being closed by server handler
Once a handler chain completes (`echo` for instance), the server's PacketConn is closed when `handle()` returns.  Xref:  https://github.com/mholt/caddy-l4/blob/master/layer4/server.go#L97 ... This means the server will no longer accept incoming requests for this listener.  I see the server's PacketConn is wrapped in a `fakeClosePacketConn` but is destructed in this case since it's the last usage.  

I'm pretty sure this a logic error for datagram protocols at least?  And apologies, but I'm too inexperienced with this framework to suggest a fix at this point.

This can be reproduced using just the echo handler: 

```json
{
  "apps": {
    "layer4": {
      "servers": {
        "test": {
          "listen": ["udp/0.0.0.0:12345"],
          "routes": [
            {
              "handle": [{"handler": "echo"}]
            }
          ]
        }
      }
    }
  }
}
```
Then send any UDP traffic to port 12345.  I use `dig @localhost -p 12345 example.com`.  The first request will be echoed back (with warnings but that's beside the point) but subsequent requests will fail.

I'm able to reproduce this with caddy-l4@master using Caddy v2.4.6 and v2.5.0-beta.1.

francislavoie:
Should fix #53, and a bunch of other issues saying that UDP wasn't working properly, that were closed without the problem being totally understood

### DIFF
--- a/layer4/server.go
+++ b/layer4/server.go
@@ -131,6 +131,11 @@ func (pc packetConn) Write(b []byte) (n int, err error) {
 	return pc.PacketConn.WriteTo(b, pc.addr)
 }
 
+func (pc packetConn) Close() error {
+	// Do nothing, we don't want to close the UDP server
+	return nil
+}
+
 func (pc packetConn) RemoteAddr() net.Addr { return pc.addr }
 
 var udpBufPool = sync.Pool{


mholt:
Thanks for the patch. Guess I didn't consider this (I don't really use the UDP sockets myself).

Does this cause a leak?

francislavoie:
Umm, no, cause it gets closed correctly here: https://github.com/mholt/caddy-l4/blob/d1f4337132468dfe2d081baba3769dfd1f46fbf8/layer4/app.go#L97

mholt (approved):
Looking into this more closely, this is probably the most correct fix. Thank you!

PR #56, opened 2022-04-11
Fix: cx.buf containing invalid content

### LINKED ISSUE #55: SNI matching after proxy_protocol handler does not work
Hi,
thanks for this awesome project, it makes very cool configurations possible.
But I think I found a bug in the way the proxy_protocol handler works. It seems not to be possible to do sni matching after the proxy_protocol handler was executed.
I would expect the following config to work, but instead I see this in the logs:
```
debug layer4.handlers.proxy_protocol  received the PROXY header       {"remote": "127.0.0.1:37055", "local": "127.0.0.1:10443"}
debug layer4.matchers.tls     matched {"remote": "127.0.0.1:37055", "server_name": "localhost"}
error    layer4  handling connection     {"error": "tls: first record does not look like a TLS handshake"}
debug layer4  connection stats        {"remote": "127.0.0.1:37055", "read": 230, "written": 0, "duration": 0.0081304}
```
<details>
<summary>caddy.json</summary>

```json
{
  "admin": {
    "disabled": true
  },
  "logging": {
    "logs": {
    }
  },
  "apps": {
    "tls": {
      "certificates": {
        "automate": ["localhost"]
      },
      "automation": {
        "policies": [{
          "subjects": ["localhost"],
          "issuers": [{
            "module": "internal"
          }]
        }]
      }
    },
    "layer4": {
      "servers": {
        "https": {
          "listen": ["0.0.0.0:10443"],
          "routes": [
            {
              "match": [
                {"proxy_protocol": {}}
              ],
              "handle": [
                {
                  "handler": "proxy_protocol",
                  "timeout": "2s",
                  "allow": ["127.0.0.1/32"]
                }
              ]
            },
            {
              "match": [
              ],
              "handle": [
                {"handler": "tls"},
                {
                  "handler": "proxy",
                }
              ]
            }
          ]
        }
      }
    },
    "http": {
      "servers": {
        "backend": {
          "allow_h2c": true,
          "routes": [
            {
              "handle": [{
                "handler": "static_response",
                "status_code": "200",
                "body": "Hello World\n",
                "headers": {
                }
              }]
            }
          ]
        }
      }
    }
  }
}
```

</details>

You should be able to run this config yourself since it only uses localhost.  
For testing I used `curl -v --insecure --haproxy-protocol https://localhost:10443`

When the match key for the second route is completely removed, so it matches all requests the above curl works and prints `Hello World`. But in my real config I can not use a match all route since I have to distribute requests to different servers.

ydylla:
Closes #55

### DIFF
--- a/layer4/connection.go
+++ b/layer4/connection.go
@@ -92,6 +92,12 @@ func (cx *Connection) Read(p []byte) (n int, err error) {
 	cx.bytesRead += uint64(n)
 
 	if !cx.recording {
+		// We read past buf and are not recording,
+		// which means cx.buf is now invalid and should be emptied for a clean next recording.
+		// Otherwise, there will be a gap in the data. Also see issue 55.
+		if cx.buf.Len() > 0 {
+			cx.buf.Reset()
+		}
 		return
 	}
 
--- a/layer4/connection.go
+++ b/layer4/connection.go
@@ -125,9 +125,13 @@ func (cx *Connection) Write(p []byte) (n int, err error) {
 // our Connection type (for example, `tls.Server()`).
 func (cx *Connection) Wrap(conn net.Conn) *Connection {
 	return &Connection{
-		Conn:    conn,
-		Context: cx.Context,
-		buf:     cx.buf,
+		Conn:         conn,
+		Context:      cx.Context,
+		buf:          cx.buf,
+		bufReader:    cx.bufReader,
+		recording:    cx.recording,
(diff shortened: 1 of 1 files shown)


mholt (approved):
This is a pretty elegant fix, thank you! Really appreciate the time and effort spent on the investigation and for the nice patch.

ydylla:
🥳 Thanks

PR #58, opened 2022-04-18
refactor: rewrite proxy protocol matcher

ydylla:
Do not use a bufio.Reader to make the fix (#56) for issue #55 more reliable.


modules/l4proxyprotocol/matcher.go
```diff
@@ -46,39 +44,15 @@ func (MatchProxyProtocol) CaddyModule() caddy.ModuleInfo {
 
 // Match returns true if the connection looks like it is using the Proxy Protocol.
 func (m MatchProxyProtocol) Match(cx *layer4.Connection) (bool, error) {
-	p := make([]byte, 4)
-	_, err := io.ReadFull(cx, p)
+	buf := make([]byte, 12) // len(headerV2Prefix)
```
RussellLuo, line 47: Maybe directly using `len(headerV2Prefix)` is better? If we adjust the header bytes in the future, although seems impossible, then there's no need to update the number again.

modules/l4proxyprotocol/handler_test.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+}
+
+func injectConfig(ctx *caddy.Context, cfg *caddy.Config) {
+	mutable := reflect.ValueOf(ctx).Elem()
+	field := mutable.FieldByName("cfg")
+	unsafePtrToField := unsafe.Pointer(field.UnsafeAddr())
+	ptrToField := (**caddy.Config)(unsafePtrToField)
+	*ptrToField = cfg
+}
```
mholt, line 29: As mentioned in the issue, I will see what I can do to make it so this is not necessary -- yikes! :)

modules/l4proxyprotocol/matcher.go
```diff
@@ -46,39 +44,15 @@ func (MatchProxyProtocol) CaddyModule() caddy.ModuleInfo {
...
-			return false, err
-		}
 
-		// 4 bytes were already read, so i == 1 is the 5th byte
-		if i != 1 {
-			return false, err
-		}
-
-		buf = append(buf, 0x00)
+	if bytes.Equal(buf[:5], headerV1Prefix) {
```
RussellLuo, line 53: To avoid the magic number, I think [bytes.HasPrefix](https://pkg.go.dev/bytes#HasPrefix) might be a better alternative.

modules/l4proxyprotocol/matcher.go
ydylla: This was premature optimization from me 😄   
Since headerV2Prefix is not a const I think the compiler can not precalculate the length, meaning it is executed on every request.  
The performance difference is probably not measurable, and I also did not verify if this is actually true.  
I changed it.

modules/l4proxyprotocol/matcher.go
ydylla: Good point

modules/l4proxyprotocol/handler_test.go
ydylla: I removed it with https://github.com/mholt/caddy-l4/pull/58/commits/51c94ed5d159160f663a49c40727c1140800c65c and instead increased the version of caddy in go.mod

modules/l4proxyprotocol/matcher.go
```diff
@@ -26,7 +24,7 @@ import (
 
 // https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt
 var (
-	headerV1Prefix = []byte("PROX") // intentional to not include "Y", see match() function for details
+	headerV1Prefix = []byte("PROXY")
```
mholt, line 27: Does this logic that was removed no longer apply?

```
// since 5th byte being null is valid in v2 header
// but is considered EOF by readers
```

Just wondering if this will break some things

ydylla: Yes it is no longer needed. I assumed it was needed because bufio has some special handling for text:

> // Package bufio implements buffered I/O. It wraps an io.Reader or io.Writer
> // object, creating another object (Reader or Writer) that also implements
> // the interface but provides buffering and some help for textual I/O.

but I did not dig deep into it.  
Instead I wrote a test to verify that io.ReadFull can read past a null byte just fine, see TestProxyProtocolMatchV2. The headerV2Prefix always contains 0x00 at the 5th position.

PR #60, opened 2022-04-20
feat: http2 support for http matcher

### LINKED ISSUE #51: [Question]: layer4 module seems cannot match http/2 or gRPC's header while caddy server does ?
The layer4 module seems cannot reconizge grpc/http2 connections with configuration below.
There's no upstream record in logs :
```
        { "handle":[{"handler":"tls"}]},
        { "handle":[
            {"handler":"proxy","upstreams":[{"dial":["127.0.0.1:8003"]}]}
          ],
          "match":[{"http":[{"header":{"Content-Type":["application/grpc*"]}}]}
          ]
        { "handle":[
            {"handler":"proxy","upstreams":[{"dial":["127.0.0.1:8003"]}]}
          ],
          "match":[{"http":[{"path":["/grpc"]}]}
          ]
        },
```
Both matchers does not work.

I only find a way to upstream gRPC now is using layer4 to proxy all connections (incl gRPC) into caddy server and then use reverse_proxy by match Content-Type or path.
example:
```
          { "handle":[{
              "handler":"reverse_proxy",
              "flush_interval":-1,
              "upstreams":[{"dial":gRPC Server"}],
              "transport":{"protocol":"http","versions":["h2c"]}}],
            "match":[{"header":{"Content-Type":["application/grpc*"]},"path":["/grpc"]}]},
```
So what's the difference between caddy server and layer4 in gRPC's header matcher ?
May I upstream gRPC connections by using layer4 module ?

ydylla:
Closes #51

### DIFF
--- a/modules/l4http/httpmatcher.go
+++ b/modules/l4http/httpmatcher.go
@@ -18,11 +18,14 @@ import (
 	"bufio"
 	"encoding/json"
 	"fmt"
-	"net/http"
-
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/modules/caddyhttp"
 	"github.com/mholt/caddy-l4/layer4"
+	"golang.org/x/net/http2"
+	"golang.org/x/net/http2/hpack"
+	"io"
+	"net/http"
+	"net/url"
 )
 
 func init() {
@@ -74,10 +77,17 @@ func (m MatchHTTP) Match(cx *layer4.Connection) (bool, error) {
 	req, ok := cx.GetVar("http_request").(*http.Request)
 	if !ok {
 		var err error
-		req, err = http.ReadRequest(bufio.NewReader(cx))
+		bufReader := bufio.NewReader(cx)
+		req, err = http.ReadRequest(bufReader)
 		if err != nil {
 			// TODO: find a way to distinguish actual errors from mismatches
 			return false, nil
+		}
+
+		// check if req is a http2 request made with prior knowledge and if so parse it
+		err = m.handleHttp2WithPriorKnowledge(bufReader, req)
+		if err != nil {
+			return false, err
 		}
 
 		// in order to use request matchers, we have to populate the request context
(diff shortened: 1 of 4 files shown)


ydylla:
@mholt I have added some http matcher tests. You can review it now.  
I am also using it since yesterday with real http2 traffic from browsers and until now I did not see any errors in the logs. So I am pretty confident I did not miss something important 😄

ydylla:
I just noticed I forgot to test the "protocol" matcher and it also does not work. Requests parsed by layer4 never have a filled TLS field, which the caddy `MatchProtocol` matcher uses. But I think the http matcher could fake one and fill it with information from the ClientHelloInfo that is saved at `cx` when the tls handler was executed before. Or at least with the server name when only the tls matcher was executed before.  
Is this a reasonable idea @mholt?

mholt:
Ah yeah, that sounds good to me. Let's try it out!

ydylla:
I will give it a try tomorrow and push it here.

ydylla:
@mholt Instead of faking the tls connection state I decided to actually use the real one. The tls handler now saves them to the layer4 connection similar to the client hellos.  

Now it's possible to use a config with a single server that listens to port 80 & 443, accepts http & https traffic on both and also redirects http traffic to use https. The single server config block is not really necessary, but it's fun 😄

ydylla:
Hi @mholt,
did you find time to review this?  
I am already using it and it works fine, so I think it can be merged.

ydylla:
Hi @mholt,
did you find time to review this?  
I am already using it and it works fine, so I think it can be merged.

mholt:
Eek, I did not yet. I was busy with the release of Caddy 2.5. sorry. I will try to look at this weekend or early next week. I think this is really cool!

modules/l4http/httpmatcher.go
```diff
@@ -96,6 +113,80 @@ func (m MatchHTTP) Match(cx *layer4.Connection) (bool, error) {
...
+	}
+
+	// According to http.Request.URL docs it only contains the value of RequestURI (so path only),
+	// but we can fill in more information
+	req.URL, err = url.Parse(fmt.Sprintf("%s://%s%s", scheme, authority, path))
+	if err != nil {
+		return err
+	}
+
+	return nil
```
mholt, line 187:
```suggestion
	req.URL, err = url.Parse(fmt.Sprintf("%s://%s%s", scheme, authority, path))
	return err
```

mholt (approved):
Very nice -- LGTM. Just one nit, but I think this is a quality change. A bit unfortunate we have to do so much of the parsing work ourselves, but oh well. Nicely done! Let's try it out.

ydylla:
Thanks for the review.  
I actually prefer the more verbose version and in this particular case it was because I commented it out and in multiple times 😄 
Should I click on merge or do you want to do it?

mholt:
@ydylla You do the honors!

PR #62, opened 2022-07-01
fix: make http2 matcher compatible with firefox

ydylla:
Firefox sends multiple (currently 6) priority frames before the headers frame is send, which means we have to increase the number of attempts to find the headers frame.

Apparently they are used as "virtual streams" see:
https://stackoverflow.com/a/36938232

Here is a screenshot of firefox starting a http2 connection:
http2-firefox

In this case it worked because the connection was proxied via cloudflare. But without this fix and cloudflare the connection matching fails with the "failed to read a http2 headers frame after 3 attempts" error.


mholt (approved):
Ah wow, I had no idea. Thanks for the patch!

ydylla:
Thanks for the review & merge 😄

PR #67, opened 2022-08-10
wip: Add replacer vars for mTLS connection details.

dekimsey:
This is based on the effort in caddyhttp's replacer but
with added Subject/Issuer details.

Also include a minor update to add the cipher_suite replacer for parity.

May be a wee bit overkill, but we actually do OU tests on incoming connections to validate which component in our stack issued the client cert. It did make me think one could refactor this to be sitting in caddy's caddytls module so both may share the Replacer with corresponding prefixes.

I added a basic test based on the http module but didn't know how to go about mocking the entire connection so some of the clientHello replacer values aren't tested.

### DIFF
--- a/modules/l4tls/handler.go
+++ b/modules/l4tls/handler.go
@@ -15,8 +15,21 @@
 package l4tls
 
 import (
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"crypto/elliptic"
+	"crypto/rsa"
+	"crypto/sha256"
 	"crypto/tls"
+	"crypto/x509"
+	"encoding/asn1"
+	"encoding/base64"
+	"encoding/pem"
 	"fmt"
+	"net"
+	"net/url"
+	"strconv"
+	"strings"
 
 	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/modules/caddytls"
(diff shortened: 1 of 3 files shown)


mholt:
Hey, thanks for the PR. Sorry I haven't replied yet. Been really busy trying to get Caddy 2.6 ready to go. This is still on my list :upside_down_face:

mholt (approved):
LGTM. Thank you for the contribution!

Anything else you want to do before I merge?

dekimsey:
I think I should remove some of those FIXMEs. What would you like me to do with them?

modules/l4tls/handler.go
```diff
@@ -104,6 +120,209 @@ func (t *Handler) Handle(cx *layer4.Connection, next layer4.Handler) error {
 	return next.Handle(cx.Wrap(tlsConn))
 }
 
+// FIXME: Should this perhaps be moved instead to caddytls?
```
mholt, line 123: Let's start by trying it here first, then it'll be easier to consider moving it upstream. :+1:

modules/l4tls/handler_test.go
```diff
@@ -0,0 +1,149 @@
...
+		NegotiatedProtocol:         "h2",
+		NegotiatedProtocolIsMutual: true,
+	}
+	addTLSVarsToReplacer(repl, cs)
+
+	for i, tc := range []struct {
+		input  string
+		expect string
+	}{
+		// FIXME: These are set during the match phase, but it's unclear how to construct a test that properly
```
mholt, line 68: I probably need to build out more test tools for this module (and Caddy too).

vnxme:
I decided to revive this PR. I rebased it on master keeping the two original commits by @dekimsey. Then I synchronised it with Caddy, added some documentation and made a few minor updates to the matcher.

I agree that testing could have been further improved by running a test server and establishing connections against it, but the current state is also OK.

dekimsey:
Thank you very much @vnxme, much appreciated!

PR #72, opened 2022-09-14
Non blocking matchers & matching timeout

ydylla:
Hi,
this is my best attempt to solve the blocking matchers problem when clients send not enough data (also discussed in #68).  
In the end it is a full rewrite of the layer4 connection. It now has a `prefetch` function which tries to load all data a client sends during connection setup. It does this in chunks of 1024 bytes up to 8 KiB and stops on the first short read.  
During matching a `ErrConsumedAllPrefetchedBytes` is returned if a matcher requests more data than currently available. If the routing code see this error it is ignored and the next matcher is tried.  

The only matcher that does not play well with this is the http matcher because `http.ReadRequest` forces us to use a `bufio.Reader`. That's why I also added a `MatchingBytes` function which allows matcher to get a view of all available bytes. The http matcher uses this to pre check if the data looks like http before calling `http.ReadRequest` and also to configure the buffer size for the `bufio.Reader` so it does not produce the `ErrConsumedAllPrefetchedBytes` error.  
It's a bit hacky and can probably be improved.

The matching timeout is implemented by setting `SetReadDeadline` before each matcher and can be configured per route.

There are still many things to do, this is just so you can take a peek.
1. I am not sure if the nested timeout config is necessary, Maybe move it up and rename it to `matcher_timeout`?
2. Should the max prefetch/matching byte size be configurable?
3. Fix the tests. `net.Pipe` behaves very differently from a real connection, it does not return short reads and so the prefetching breaks.
4. And probably write some more tests 😄

### DIFF
--- a/layer4/connection.go
+++ b/layer4/connection.go
@@ -15,9 +15,8 @@
 package layer4
 
 import (
-	"bytes"
 	"context"
-	"io"
+	"errors"
 	"net"
 	"sync"
 
@@ -30,7 +29,7 @@ import (
 // and variable table. This function is intended for use at the start of a
 // connection handler chain where the underlying connection is not yet a layer4
 // Connection value.
-func WrapConnection(underlying net.Conn, buf *bytes.Buffer, logger *zap.Logger) *Connection {
+func WrapConnection(underlying net.Conn, buf []byte, logger *zap.Logger) *Connection {
 	repl := caddy.NewReplacer()
 	repl.Set("l4.conn.remote_addr", underlying.RemoteAddr())
 	repl.Set("l4.conn.local_addr", underlying.LocalAddr())
@@ -66,31 +65,40 @@ type Connection struct {
 
 	Logger *zap.Logger
 
-	buf       *bytes.Buffer // stores recordings
-	bufReader io.Reader     // used to read buf so it doesn't discard bytes
+	buf       []byte // stores recorded data
+	offset    int
 	recording bool
 
 	bytesRead, bytesWritten uint64
 }
+
+var ErrConsumedAllPrefetchedBytes = errors.New("consumed all prefetched bytes")
+var ErrMatchingBufferFull = errors.New("matching buffer is full")
 
 // Read implements io.Reader in such a way that reads first
 // deplete any associated buffer from the prior recording,
 // and once depleted (or if there isn't one), it continues
 // reading from the underlying connection.
 func (cx *Connection) Read(p []byte) (n int, err error) {
+	if cx.recording {
+		if len(cx.buf) == 0 {
+			err = cx.prefetch()
+			if err != nil {
+				return 0, err
+			}
+		}
+		if len(cx.buf[cx.offset:])-len(p) < 0 {
+			return 0, ErrConsumedAllPrefetchedBytes
+		}
+	}
+
 	// if there is a buffer we should read from, start
 	// with that; we only read from the underlying conn
 	// after the buffer has been "depleted"
-	if cx.bufReader != nil {
-		n, err = cx.bufReader.Read(p)
-		if err == io.EOF {
-			cx.bufReader = nil
-			err = nil
-		}
-		// prevent first read from returning 0 bytes because of empty bufReader
-		if !(n == 0 && err == nil) {
-			return
-		}
+	if cx.offset < len(cx.buf) {
+		n := copy(p, cx.buf[cx.offset:])
+		cx.offset += n
+		return n, nil
 	}
 
 	// buffer has been "depleted" so read from
@@ -98,16 +106,11 @@ func (cx *Connection) Read(p []byte) (n int, err error) {
 	n, err = cx.Conn.Read(p)
 	cx.bytesRead += uint64(n)
 
-	if !cx.recording {
-		return
-	}
-
-	// since we're recording at this point, anything that
-	// was read needs to be written to the buffer, even
-	// if there was an error
-	if n > 0 {
-		if nw, errw := cx.buf.Write(p[:n]); errw != nil {
-			return nw, errw
+	if cx.recording {
+		cx.buf = append(cx.buf, p[:n]...)
+		cx.offset += n
+		if len(cx.buf) >= MaxMatchingBytes {
+			return n, ErrMatchingBufferFull
 		}
 	}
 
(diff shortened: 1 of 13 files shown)


mholt:
Can't wait to review this!

Please hold tight while I get Caddy 2.6 released. Should be this week if all goes well.

mholt:
Ok, now I'm working through my backlog from the 2.6 release, so, hang tight :grinning:

ydylla:
Just played a bit more with this. The tests are now fixed and simple configs should already work.  
Sadly http2 matching is broken. The first prefetch only receives the http1 upgrade request but no http2 frames which it needs for matching (tested with curl, browsers may behave different).  
I think it is fixable but it gets uglier and uglier 😢

mholt:
@ydylla Thanks for working on this! I'm still backlogged and have been taking a little time for my mental health lately but I will be back to this as soon as I can :)

I sympathize with the complexity of this... I hope we can eventually solve these in an efficient way. Really appreciate your contributions :100:

layer4/connection.go
```diff
@@ -15,21 +15,21 @@
...
+	"go.uber.org/zap"
 )
 
 // WrapConnection wraps an underlying connection into a layer4 connection that
 // supports recording and rewinding, as well as adding context with a replacer
 // and variable table. This function is intended for use at the start of a
 // connection handler chain where the underlying connection is not yet a layer4
 // Connection value.
-func WrapConnection(underlying net.Conn, buf *bytes.Buffer) *Connection {
+func WrapConnection(underlying net.Conn, buf []byte, logger *zap.Logger) *Connection {
```
mholt, line 32: What's wrong with `*bytes.Buffer`, exactly?

I'm also curious what the need is for a different logger on each connection?

Thanks!

mholt (commented):
I finally had a chance to look at this! This is really impressive, and overall I think I like where this is going. Bear with me as I have some questions though, as I haven't looked at this code in quite a while! So don't mind my questions -- they are not criticisms -- I am just trying to understand this as well as you do.

> It now has a prefetch function which tries to load all data a client sends during connection setup. It does this in chunks of 1024 bytes up to 8 KiB and stops on the first short read.

What's the advantage of this over simply reading bytes as the matchers need them (as we currently do)? ... [several minutes later] ... thinking on it, is it because we might as well read all the bytes (with a cap) the client sends to establish the connection, since reading all the bytes is what the server has to do anyways? Might as well do it all at once, I guess?

> During matching a ErrConsumedAllPrefetchedBytes is returned if a matcher requests more data than currently available. If the routing code see this error it is ignored and the next matcher is tried.

Ok, I think I get this part. The client only sends so many bytes at the beginning of a connection (within a deadline) and those are what the matcher has to work with, period.

> The only matcher that does not play well with this is the http matcher because http.ReadRequest forces us to use a bufio.Reader.

> It's a bit hacky and can probably be improved.

That does sound a bit complicated. Let's collaborate on this and see if we can come up with something simpler.

> The matching timeout is implemented by setting SetReadDeadline before each matcher and can be configured per route.

I wonder if this should be set between each Read()? I think usually it is conventional to enforce read timeouts by extending the deadline after each read.

> I am not sure if the nested timeout config is necessary, Maybe move it up and rename it to matcher_timeout?

I like where it is for now, let's see how people use it.

> Should the max prefetch/matching byte size be configurable?

Almost certainly, yes. Not a showstopper but I think that will be a good idea.

Thanks @ydylla -- hopefully you haven't given up on me yet with how long I'm taking :upside_down_face:

ydylla:
Thanks for the feedback, I will try to answer your questions.

> What's the advantage of this over simply reading bytes as the matchers need them

The main reason for the chunking is the detection of short reads. If we ask `Read` for x bytes but it returns less than x bytes we can be pretty sure that the next `Read` call will block. This would allow us to stop reading without the need of a timeout (ReadDeadline) which would always delay the connection matching (even if it's only for milliseconds).
At least I thought that this is always the case until I noticed it is not quite true for http2 for example. In reality it also seems to depend on how the client sends the data. Curl for example seems to send the http2 upgrade request first and after that the http2 frames. Instead of both together with the same "flush".  This means we get a short read but the next read could be a full read again.  
Also yes the server has to read the bytes anyway.

Regarding http.ReadRequest and bufio.Reader. Yes a way to parse http request without the need of bufio would be nice. We already have all bytes in memory so it's unnecessary.  
But right now I really do not see a way, maybe with another library.

> I wonder if this should be set between each Read()?

The current timeout was only intended as a total matching timeout, not a read timeout. If no route could be matched in x seconds it's very likely not a legitimate client, so the server should close the connection. After a route is selected `Read`'s are allowed to take longer (indefinitely) like before. But I agree a general read timeout would also be a good idea. Then the deadline has to be refreshed after each `Read`.

Regarding the max prefetch size. I was mostly unsure how to access config that is on server level from the connection.
Either the config value has to be duplicated into the connection or connections need a reference to their server.

layer4/connection.go
ydylla: I found it a lot easier to manage and think about the offset myself instead of recreating a new Buffer each time we need to seek backwards.

ydylla:
> This is looking good I think. Did you mean to leave it in draft state?

Thanks. Yes this is still a draft because http2 matching is not working in all cases. It depends on how the client sends the data. See my last comment:
> I noticed it is not quite true for http2 for example. In reality it also seems to depend on how the client sends the data. Curl for example seems to send the http2 upgrade request first and after that the http2 frames. Instead of both together with the same "flush". This means we get a short read but the next read could be a full read again.

I also did not find the time to test it with real web browser traffic, which I planned to do before undrafting it.

mholt:
@ydylla Any interest in finishing this up? Let me know if I can help.

mholt:
Hi @ydylla -- I might actually merge this sooner rather than later, and see if we can figure out the HTTP/2 part in a separate PR, or if you want we can finish up the HTTP/2 tweaks in this one. Up to you; I know you're very busy. (I am too!) If I don't hear from you next time I circle back to this I'll probably just go forward with the merge. :smiley:

ydylla:
@mholt Sorry apparently I forgot to answer to your previous message 😅   
I don't think this can be merged like it is. It may look good but will probably not work in many cases. I think a better approach would be a global matching timeout (not per route like now) and a loop that keeps trying all routes/matchers until the timeout is reached. So the flow would be:
* prefetch until first short read
* try matchers
* if nothing matched prefetch again, which will either add more data (to the next short read) or block until the timeout is reached

I will experiment with this within the next couple of days and report back to you.

ydylla:
@mholt I completely rewrote it and opened a new PR, see https://github.com/mholt/caddy-l4/pull/192

PR #73, opened 2022-09-19
fix: shutdown half-closed connections for more connection types

ydylla:
Hi,
I noticed `net.UnixConn` upstream connections are not closed properly if the downstream goes away (sends fin-ack). Instead layer4 waits until the upstream starts to tear down the connection.  
For `net.TCPConn` upstreams it works correctly thanks to the `CloseWrite` call in [proxy.go#L286](https://github.com/mholt/caddy-l4/blob/master/modules/l4proxy/proxy.go#L286).  
This PR enables the `CloseWrite` call for more connection types.

You can reproduce it with the following config and `curl http://localhost:8080`:
<details>
<summary>close.json</summary>

```json
{
  "logging": {
    "logs": {
      "default": {"level":"DEBUG"}
    }
  },
  "apps": {
    "layer4": {
      "servers": {
        "http+https": {
          "listen": ["0.0.0.0:8080"],
          "routes": [
            {
              "match": [
                {"http": [{"host": ["localhost"]}]}
              ],
              "handle": [
                {
                  "handler": "proxy",
                  "upstreams": [{"dial": ["unix/sock"]}]
                }
              ]
            }
          ]
        }
      }
    },
    "http": {
      "servers": {
        "backend": {
          "listen": ["unix/sock"],
          "idle_timeout": "30s",
          "routes": [
            {
              "handle": [{
                "handler": "static_response",
                "status_code": "200",
                "body": "Hello World\n",
                "headers": {
                }
              }]
            }
          ]
        }
      }
    }
  }
}
```

</details>

Here is also a wireshark capture showing that layer4 ignores the fin-ack received from curl and starts doing keep alive, instead of also sending a fin-ack and closing the connection immediately. It only happens after 30s because that's the configured idle timeout of the hello world server.
[close-broken.zip](https://github.com/mholt/caddy-l4/files/9600824/close-broken.zip)

### DIFF
--- a/modules/l4proxy/proxy.go
+++ b/modules/l4proxy/proxy.go
@@ -281,8 +281,7 @@ func (h *Handler) proxy(down *layer4.Connection, upConns []net.Conn) {
 		// Shut down the writing side of all upstream connections, in case
 		// that the downstream connection is half closed. (issue #40)
 		for _, up := range upConns {
-			// only doable for a TCP connection
-			if conn, ok := up.(*net.TCPConn); ok {
+			if conn, ok := up.(closeWriter); ok {
 				_ = conn.CloseWrite()
 			}
 		}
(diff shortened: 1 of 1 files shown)


mholt (approved):
Ohh, excellent catch. Thank you for this patch!

ydylla:
Thanks for the merge 🥳

PR #74, opened 2022-10-07
fix: use caddy remote module (instead of local one)

rpoisel:
The referenced tag does not exist (anymore). This should fix this. Thanks!


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.18
 replace github.com/caddyserver/caddy/v2 => ../caddyserver/caddy
```
mohammed90, line 5: This is the fix

```suggestion
```

mohammed90 (changes requested):
Thank you for the PR, but the referenced commit actually exists. The `replace` directive is not right to stay there.

go.mod
rpoisel: Thanks for the hint and your quick response. I added two commits which in total only change what you requested.

mohammed90: Great. Can you please run `go mod tidy`? There are changes to `go.sum` to be applied by the `go mod tidy` run

mholt (approved):
Huh, oops. I can't remember how this slipped in there, but there we have it. Thanks!

rpoisel:
Thanks for your help and your patience :-) I also "fixed" the subject of this PR.

mohammed90 (approved):
Perfect! Thank you so much

mholt:
Thank you both :D

PR #85, opened 2022-11-30
introduce the CI

IceCodeNew:
in order to make sure later commits will not break the compilation

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,42 @@
+# This workflow will build a golang project
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go
+
+name: Go
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Set up Go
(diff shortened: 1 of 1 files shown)


ydylla:
Good idea.
I think the pipeline should also run all tests. You can do this with `go test ./...` this would also have caught the build failure.  

And for the xcaddy build the pipeline should probably use a replacement `--with` to ensure it actually uses the local version of layer4 from the pipeline. You can do this by adding `=.` to the `--with` option (`--with github.com/mholt/caddy-l4=.`).

IceCodeNew:
> Good idea. I think the pipeline should also run all tests. You can do this with `go test ./...` this would also have caught the build failure.
> 
> And for the xcaddy build the pipeline should probably use a replacement `--with` to ensure it actually uses the local version of layer4 from the pipeline. You can do this by adding `=.` to the `--with` option (`--with github.com/mholt/caddy-l4=.`).

Thanks for taking the time to review this PR. I have implemented the amendment ;-)

ydylla:
@IceCodeNew I have fixed the failing test in #86. So I think you can include the tests again, once @mholt merged it.

mholt (approved):
Awesome!! Thanks for doing this. I admit, I don't have a lot of CI knowledge (most of my energy is focused on the actual code, rather than the releasing of it -- I should probably learn some things though). So I really appreciate this. Let's give this a try.

PR #86, opened 2022-11-30
test: improve socks5 handler test

ydylla:
Now it's compatible with Linux and also tests the success case.  
@IceCodeNew noticed in #85 that it does not work in the pipeline.

I developed the test under Windows where the error always is "host unreachable" when trying to connect to a closed port.
But with Linux the error is "connection refused", apparently nobody has executed the test until now with Linux 🤷🏻

### DIFF
--- a/modules/l4socks/socks5_handler_test.go
+++ b/modules/l4socks/socks5_handler_test.go
@@ -3,8 +3,10 @@ package l4socks
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
 	"io"
 	"net"
+	"strconv"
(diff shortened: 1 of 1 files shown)


mholt:
Thanks for this!

I'm home taking care of sick kids today, so I'll look at this soon :) (and the other issues/PRs)

mholt (approved):
Thank you for this :smiley:

It's been a while since I worked on this repo myself, so it's been a while since I ran the tests. I'm glad we have some CI now and thanks for helping to maintain this! I hope it continues to be useful to you!

PR #88, opened 2022-12-02
Add HTTP CONNECT Handler

kkroo:
* Hijacks request connection and sends it to the next handler


modules/l4proxyconnect/handler.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/mholt/caddy-l4/layer4"
+	"go.uber.org/zap"
+	"net"
+	"net/http"
+)
+
+func init() {
+	caddy.RegisterModule(Handler{})
+}
+
```
francislavoie, line 31: Needs an update

mholt:
Thanks! This looks cool. Do you have an example of how this would be used?

And does this work with the current proxy module? Should it, I wonder?

mholt:
@kkroo Any interest in having this merged?

PR #95, opened 2023-01-31
Fix #94

### LINKED ISSUE #94: "selection_policies.ip_hash" should use same hashing algorithm as caddy
In caddy, "ip_hash" uses the "Highest Random Weight" algorithm, but not in l4, which leads to the fact that the load balancing service used for both tcp layer 7 and layer 4 cannot obtain a consistent "upstream" through "ip_hash" ", which is **not** very good for some **stateful services** that require tcp layer 7 and layer 4 at the same time.

```go
// algorithm code in caddy

// hostByHashing returns an available host from pool based on a hashable string s.
func hostByHashing(pool []*Upstream, s string) *Upstream {
	// Highest Random Weight (HRW, or "Rendezvous") hashing,
	// guarantees stability when the list of upstreams changes;
	// see https://medium.com/i0exception/rendezvous-hashing-8c00e2fb58b0,
	// https://randorithms.com/2020/12/26/rendezvous-hashing.html,
	// and https://en.wikipedia.org/wiki/Rendezvous_hashing.
	var highestHash uint32
	var upstream *Upstream
	for _, up := range pool {
		if !up.Available() {
			continue
		}
		h := hash(up.String() + s) // important to hash key and server together
		if h > highestHash {
			highestHash = h
			upstream = up
		}
	}
	return upstream
}
```

```go
// algorithm code in l4

// hostByHashing returns an available host
// from pool based on a hashable string s.
func hostByHashing(pool []*Upstream, s string) *Upstream {
	poolLen := uint32(len(pool))
	if poolLen == 0 {
		return nil
	}
	index := hash(s) % poolLen
	for i := uint32(0); i < poolLen; i++ {
		index += i
		upstream := pool[index%poolLen]
		if upstream.available() {
			return upstream
		}
	}
	return nil
}
```

see [selectionpolicies.go](https://github.com/caddyserver/caddy/blob/7f2a93e6c36ceec9e007baeedeb594ab0dc43a71/modules/caddyhttp/reverseproxy/selectionpolicies.go#L519) and [loadbalancing.go](https://github.com/mholt/caddy-l4/blob/121f2ada24819cbb71e2aab1f6ab89e99d85fcf6/modules/l4proxy/loadbalancing.go#L319)

since1986:
see [https://github.com/mholt/caddy-l4/issues/94#issuecomment-1407938527](https://github.com/mholt/caddy-l4/issues/94#issuecomment-1407938527) and [https://github.com/mholt/caddy-l4/issues/94#issuecomment-1408224103](https://github.com/mholt/caddy-l4/issues/94#issuecomment-1408224103)

### DIFF
--- a/modules/l4proxy/loadbalancing.go
+++ b/modules/l4proxy/loadbalancing.go
@@ -317,25 +317,26 @@ func leastConns(upstreams []*Upstream) *Upstream {
 // hostByHashing returns an available host
 // from pool based on a hashable string s.
 func hostByHashing(pool []*Upstream, s string) *Upstream {
-	poolLen := uint32(len(pool))
-	if poolLen == 0 {
-		return nil
-	}
-	index := hash(s) % poolLen
-	for i := uint32(0); i < poolLen; i++ {
-		index += i
-		upstream := pool[index%poolLen]
-		if upstream.available() {
-			return upstream
+	// HRW hash (copy from caddy's code)
+	var highestHash uint32
+	var upstream *Upstream
+	for _, up := range pool {
+		if !up.available() {
+			continue
+		}
+		h := hash(up.String() + s) // important to hash key and server together
+		if h > highestHash {
+			highestHash = h
+			upstream = up
 		}
 	}
-	return nil
+	return upstream
 }
 
 // hash calculates a fast hash based on s.
 func hash(s string) uint32 {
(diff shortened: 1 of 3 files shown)


mholt (approved):
Awesome, thank you for the contribution! This looks like a good port. I appreciate the test too.

PR #100, opened 2023-03-13
add local IP matcher

### LINKED ISSUE #99: No apparent way to match on local_ip
Hello, I need to distinguish routes depending on what IP Caddy's socket is created. Either I'm blind or there is no support for this condition anywhere in Caddy, but I need this specifically in L4 module so the issue goes here.

Just to give context, if you listen on 0.0.0.0, the listener socket has address of 0.0.0.0, however each actual transport socket will have specific addresses on the server side, depending on the routing table. Creating separate servers each listening on a specific address isn't a viable workaround in my case.

To reiterate, I need this logic:
```json
{
  "apps": {
    "layer4": {
      "servers": {
        "server0": {
          "listen": [
            ":1234"
          ],
          "routes": [
            {
              "match": [
                {
                  "local_ip": {
                    "ranges": [
                      "10.0.1.0/24"
                    ]
                  }
                }
              ],
              "handle": [
                " < some handlers > "
              ]
            },
            {
              "match": [
                {
                  "local_ip": {
                    "ranges": [
                      "10.0.2.0/24"
                    ]
                  }
                }
              ],
              "handle": [
                " < some other handlers > "
              ]
            }
          ]
        }
      }
    }
  }
}
```

mohammed90:
Closes #99


IndeedNotJames (changes requested):
Haven't looked at the code yet, but you probably want to add this new matcher to the list of matchers in the [README.md](https://github.com/mholt/caddy-l4/blob/58d19ae52d601387bbe7b247125fbfe0980748cb/README.md#introduction) as well :)

mholt (approved):
Brilliant! You're fast. Thanks Mohammed!

mholt:
Oops, I need to update the CI to use Go 1.20.

PR #105, opened 2023-03-22
ci: version-ctrl

IceCodeNew:
test against the go version used by the upstream
& the next version of that

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -45,7 +45,7 @@ jobs:
         go-version-file: ./go.mod
 
     - name: Set up xcaddy
-      uses: engineerd/configurator@v0.0.8
+      uses: engineerd/configurator@v0.0.10
       with:
         name: ${{ matrix.name }}
         pathInArchive: ${{ matrix.pathInArchive }}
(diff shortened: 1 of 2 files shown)


IceCodeNew:
@mholt it seems that the [dependabot](https://github.com/apps/dependabot) is not enabled in this repo. I am wondering if we can let that bot watching the dependencies update for us.
If you are OK with it, please configure the bot to watch both the dependencies of GitHub Action & Go.

francislavoie:
I don't think we really want to do that, we want to keep the go.mod set to the minimum version we support. Really it's now 1.19 as of https://github.com/caddyserver/caddy/commit/8cb1bb4af30b880f981b1ae6fdb13e5944bfaefe.

IceCodeNew:
> I don't think we really want to do that, we want to keep the go.mod set to the minimum version we support. Really it's now 1.19 as of [caddyserver/caddy@8cb1bb4](https://github.com/caddyserver/caddy/commit/8cb1bb4af30b880f981b1ae6fdb13e5944bfaefe).

Fair enough, let's keep on maintaining the matrix of go versions ;-)

francislavoie (approved):
LGTM

francislavoie:
Re dependabot, you could pull this in https://github.com/caddyserver/caddy/blob/master/.github/dependabot.yml

PR #133, opened 2023-08-01
fix typo in json field name

daniel-naegele:


mohammed90 (approved):
Thanks for catching this!

PR #153, opened 2023-10-16
Bump versions

IceCodeNew:


IceCodeNew:
Sorry for the force push on the master branch, I had pushed to the master branch directly. Considering all of the options, I decided to falsify the history.

BTW please consider protecting the master branch ;-)

.github/workflows/go.yml
```diff
@@ -58,7 +60,7 @@ jobs:
 
     - name: Build
       run: |
-        xcaddy build master --output ${{ runner.temp }}/${{ matrix.name }} --with github.com/mholt/caddy-l4=./
+        xcaddy build latest --output ${{ runner.temp }}/${{ matrix.name }} --with github.com/mholt/caddy-l4=./
```
francislavoie, line 63: I don't think this is right, `latest` isn't a git ref that exists.

mohammed90: `latest` is a special case of the `go {get,mod}` resolver which points to the latest tagged release. This is specified here:
https://go.dev/ref/mod#version-queries

IceCodeNew: Thanks for pointing that out, I was narrowing down what is causing the CI failed. So the change here won't get into the final PR.

BTW I believe the `latest` would be recognized as a keyword, refering to https://github.com/caddyserver/xcaddy/blob/f42cc51e4bad735f1e23616109012b56217c72f0/README.md?plain=1#L68

IceCodeNew: @francislavoie @mohammed90 Thanks for taking the time to review this PR. This change of the `xcaddy build latest` instruction has been reverted.

francislavoie: Oh right, my bad, forgot we had that :+1:

francislavoie (approved):

Issue #119, opened 2023-05-08
Tunnelling legacy TLS1.0 clients to another host, terminating modern clients


agittins:
I have caddy running in a docker container and reverse-proxying a variety of services and it's been working great for months.

I now need to host a client's legacy service, which has clients with old libs and hard-coded certs (they're not browsers). The clients only support TLS1.0(!) and I only have a single IPv4 address on which to serve both the legacy clients and my modern services.

I am not able to have the clients upgraded at this time, so my job is to just make it work until the business side of things can resolve the older clients out in the field. I know that caddy doesn't support tls1.0 so I figure a non-terminating pass-through sort of solution might be the best answer (out of a collection of bad answers).

For now I've dnat'd all port 443 traffic to the legacy server which works but of course that means my other services are now down.

Is caddy-l4 a possible solution to my problem? In theory, the legacy clients could be identified at the TLS Client Hello packet, as they specify TLS1.0 and it would be suitable to proxy/tunnel the session to my legacy upstream and let it take care of termination etc. Any other clients (who are not using an ancient, weak implementation) would go through the normal SNI process, be locally TLS terminated and reverse-proxied to their destinations as usual.

Despite being an awful situation, am I on the right track? Is there an easier / better way to support this? My firewall is using nftables if that's relevant. I am at a bit of a loss as to how I'd configure caddy-l4 to do this so I'm not sure if it's outside of what caddy-l4 can do or I'm just being daft :-)

Any ideas (within the operational constraints I have) welcome!

agittins:
(I suspect my issue is perhaps somewhat similar to #70 but perhaps further complicated by having to support a tls version that caddy doesn't implement).

mholt:
Although Caddy's TLS app doesn't support TLS < 1.2 anymore, if you're not terminating TLS then that app won't be used, and you should be able to match TLS 1.0 connections. Though I haven't tested it.

agittins:
This is fixed for me in #155 

Discussion etc at https://caddy.community/t/how-do-i-match-tls-version-in-layer4/21583/9

vnxme:
With what's added in #396, we don't need a separate TLS version matcher to route old clients away -- it's now possible to handle it using placeholders like this:
```caddyfile
{
	layer4 {
		tcp/:4443 {
			@tls tls
			route @tls {
				subroute {
					# detect TLS 1.0 and 1.1 clients
					# and proxy to insecure.local:443
[...]

PR #155, opened 2023-11-04
Add tls.handshake_match.version matcher

agittins:
- creates a new matcher, tls.handshake_match.version
- based heavily on the alpn matcher.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Current matchers:
 - **layer4.matchers.socks4** - matches connections that look like [SOCKSv4](https://www.openssh.com/txt/socks4.protocol).
 - **layer4.matchers.socks5** - matches connections that look like [SOCKSv5](https://www.rfc-editor.org/rfc/rfc1928.html).
 - **tls.handshake_match.version** - matches TLS Version from ClientHello, using strings from [VersionName](https://go.dev/src/crypto/tls/common.go).
+
 Current handlers:
 
 - **layer4.handlers.echo** - An echo server.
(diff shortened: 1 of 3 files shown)


agittins:
Fixes #119

mholt (commented):
Thanks for this!

Since the TLS handshake hasn't been completed yet, a version hasn't actually been chosen. This only matches versions the client advertises support for. Is that intentional? If so, we should probably update the godoc to clarify.

Normally I suppose we'd just put this in the `tls` app package directly, but that's not strictly required and this is the only request I've had for it, so maybe keeping it in this separate module is better for now.

agittins:
That's a good point - the idea is to match against versions the client supports (and especially those which the tls app doess *not* support), but it should probably be more specific - in my case I have clients that *only* support TLS 1.0, so that's what I was looking for - but yes the more general case is that clients offer a max version or a list.

I think even for my use-case I should be checking for "clients that support a max_version of tls1.0", since I don't want to redirect old browsers etc that support newer versions but still offer support of 1.0.

I might give this a bit more thought but if you have time to offer some insight it would be more than welcome.

mholt:
Huh, that's super weird. I admit I don't know. Go doesn't support SSLv2; maybe our parser will still decode it, but I'd be surprised if it worked.

vnxme:
@agittins thank you contributing!

What this matcher is intended to do is now covered by the `vars` matcher and the `tls` matcher exposing the `{l4.tls.version}` placeholder. See also [this comment](https://github.com/mholt/caddy-l4/issues/119#issuecomment-4047589421).

Closing this PR as not planned.

PR #164, opened 2023-12-02
matchers: implement `not`

### LINKED ISSUE #160: Add `not` Matcher
Matching negation could be very useful. It's very simple to implement. The contributor could replicate the logic from the Caddy HTTP app here:

https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/matchers.go#L1187-L1245

mohammed90:
Closes #160

### DIFF
--- a/layer4/matchers.go
+++ b/layer4/matchers.go
@@ -15,6 +15,7 @@
 package layer4
 
 import (
+	"encoding/json"
 	"fmt"
 	"net"
 	"strings"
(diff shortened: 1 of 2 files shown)


mholt (approved):
Given my large backlog, I admit I only did a cursory look at the code, but overall this looks good! Especially where there's already a comparable implementation that this is based on. Thanks for contributing :smiley:

PR #165, opened 2023-12-03
matchers: use `netip`

mohammed90:
To be merged after #164


mholt:
This is looking good to me so far. We just need to rebase with master now.

PR #186, opened 2024-05-02
L4 module for Postgres

metafeather:
This a simple module to detect and match on Postgres connections in a similar manner that the SSH module uses.

Currently it can be used to multiplex a single database connection along with HTTP/S and SSH to avoid needed to open additional ports:

```json
            "routes": [
              {
                "match": [
                  {
                    "postgres": {}
                  }
                ],
                "handle": [
                  {
                    "handler": "proxy",
                    "upstreams": [
                      {
                        "dial": [
                          "postgres.localhost:5432"
                        ]
                      }
                    ]
                  }
                ]
              }
            ]
```

### DIFF
--- a/modules/l4postgres/matcher.go
+++ b/modules/l4postgres/matcher.go
@@ -1,4 +1,18 @@
-// Allows the L4 multiplexing of Postgres connections
+// Copyright 2020 Matthew Holt
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package l4postgres allows the L4 multiplexing of Postgres connections
 //
 // With thanks to docs and code published at these links:
 // ref: https://github.com/mholt/caddy-l4/blob/master/modules/l4ssh/matcher.go
@@ -7,7 +21,6 @@
 // ref: https://ivdl.co.za/2024/03/02/pretending-to-be-postgresql-part-one-1/
 // ref: https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-STARTUPMESSAGE
 // ref: https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-SSLREQUEST
-
 package l4postgres
 
 import (
--- a/modules/l4postgres/matcher.go
+++ b/modules/l4postgres/matcher.go
@@ -37,24 +37,26 @@ func init() {
 }
 
 const (
-	SSLRequestCode = 80877103
-	InitMessageSizeLength = 4
+	// Magic number to identify a SSLRequest message
+	sslRequestCode = 80877103
+	// byte size of the 
+	initMessageSizeLength = 4
 )
 
 // Message provides readers for various types and 
 // updates the offset after each read
-type Message struct{
+type message struct{
 	data []byte
 	offset uint32
 }
 
-func (b *Message) ReadUint32() (r uint32) {
+func (b *message) ReadUint32() (r uint32) {
 	r = binary.BigEndian.Uint32(b.data[b.offset : b.offset+4])
 	b.offset += 4
 	return r
 }
 
-func (b *Message) ReadString() (r string) {
+func (b *message) ReadString() (r string) {
 	end := b.offset
 	max := uint32(len(b.data))
 	for ; end != max && b.data[end] != 0; end++ {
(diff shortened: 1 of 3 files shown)


metafeather:
Thanks for the feedback, all done

modules/l4postgres/matcher.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+func init() {
+	caddy.RegisterModule(MatchPostgres{})
+}
+
+const (
+	// Magic number to identify a SSLRequest message
+	sslRequestCode = 80877103
+	// byte size of the
```
mholt, line 42: Of the what !? The suspense is killing me!

mholt (approved):
Thank you! Looks good for now. Let's give it a try :)

modules/l4postgres/matcher.go
metafeather: Literally just a magic number picked by Postgres for the protocol: https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-SSLREQUEST

metafeather: I wonder if we should add a note to review when Postgres versions approach v1234.5679 🤔 (and why 5679 not 5678?)

mholt: I understand about the `sslRequestCode = 80877103` -- I was commenting on the line after with an incomplete sentence: `// byte size of the `

metafeather: Doh, I just came across this myself when looking back through the code and realised I'd misunderstood 🙈

mholt: Ah, haha, no problem. Thanks for the fix!

mholt (approved):

coolaj86:
The first 8 bytes are the Postgres' bespoke `SSLRequest`.

The server (`caddy`) needs to respond with `S`, after which the very next bytes will be the normal TLS hello containing the SNI so that the connection can be matched and routed to the correct server.

At that time, the `SSLRequest` needs to be sent to the Postgres Server, the `S` response byte should be eaten, and the connection can handed off completely at that point.

**Without sending `S`, there's no way to know which Postgres server the request should be sent to.**

It looks like we could add the `cx.Write('S')`... but I'm not clear on how we'd pass that off to the TLS handler and then hand control back for the `SSLRequest` to be repeated before letting go of control.

modules/l4postgres/matcher.go
```diff
@@ -0,0 +1,133 @@
...
+	data := make([]byte, binary.BigEndian.Uint32(head)-initMessageSizeLength)
+	if _, err := io.ReadFull(cx, data); err != nil {
+		return false, err
+	}
+
+	b := newMessageFromBytes(data)
+
+	// Check if it is a SSLRequest
+	code := b.ReadUint32()
+	if code == sslRequestCode {
```
coolaj86, line 109: I think `cx.Write('S')` would go here, but then how to apply a TLS matcher?

mohammed90: A matcher is the wrong place to write outputs. Writings go into handlers.

coolaj86: How do you propose to finish matching and pass to a handler if you don't have the bytes to match yet?

Or are you suggesting that the rest of the matching be completed in a handler?

mohammed90: There would be another set of matcher/handler chain, akin to the [subroute handler](https://caddyserver.com/docs/json/apps/http/servers/routes/handle/subroute/) that's available in the `http` app. Assuming the existence of a `subroute` handler and a `writer` handler for the convenience of the discussion, it'll look something like this:

```json
{
	"routes": [
		{
			"match": [
				{
					"postgres": {}
				}
			],
			"handle": [
				{
					"handler": "writer",
					"bytes": "S"
				},
				{
					"handler": "subroute",
					"routes": [
						{
							"match": [
								{
									"tls": {
										"sni": [
											"postgres.example.net"
										]
									}
								}
							],
							"handle": [
								{
									"handle": "tls"
								},
								{
									"handle": "proxy",
									"upstreams": [
										{
											"dial": "postgres.localhost:5432"
										}
									]
								}
							]
						}
					]
				}
			]
		}
	]
}
```

coolaj86: The abstraction has to be a bit higher than that, otherwise what you'll end up with is a very tedious programming language for which the AST is literally this JSON structure of matchers and handlers.

That said, I kinda get what you're driving at.

To convey the idea without writing out all the syntax:

```
matcher: postgres
  - handler: postgres-ssl
    # encrypted connection
    - matcher: tls
      - sni: customer1.example.com
        - handler: postgres-tls-proxy
    # plain connection
    - matcher: postgres-connection
      - hostname: customer1.example.com
        - handler: proxy-plain-proxy
```

You couldn't use the normal `proxy` because you have to replay the SSLRequest ``, eat the `S`, and then resume the connection.

This is so highly specific to postgres... just seems like it'll be a real pain to break it down into generic abstractions just to try to fit into a pattern that was created without the foresight of this type of interaction.

coolaj86: Thinking on it a bit more: This is very much *NOT* a generic flow. There are a finite number of outcomes:
1. Plain Postgres TCP (`sslmode=disable`) - match on DB and proxy
2. SSLRequest (`sslmode=default|require|verify-partial|verify-full`) \
   no matching info available yet, so the config must indicate whether to send `S` to the client... or whatever byte means "not encrypted"
   - Proceed to 1 (above) if the config is not encrypted
   - The connection is encrypted, so there are now 2 options:
       1. terminate TLS now, then proceed to 1 (above)
       2. match on TLS, then replay the client connection, then proxy

mohammed90: At this point, you should move this conversation and conclusion to #187. @metafeather may have better reasoning over this than I because he's working with this.

Issue #187, opened 2024-05-04
How to match on StartTLS for proxying Postgres?


coolaj86:
## Update

`psql` will send `00 00 00 08 04 d2 16 2f` (SSLRequest) and wait for either
- `N` (no - plaintext)
- `Y` (yes - version 2) (deprecated)
- or `S` (ssl - version 3)

If `S` is sent, it will immediately begin a standard TLS connection.

After TLS is terminated, of if `N` (plaintext) is sent, it will send `00 00 xx xx 00 03 00 00` - where xx are Little-Endian length bytes (up to 4k at least) and 03 means "pg version 3", followed by the plaintext db connection information (username, dbname, application name, character encoding).

If `sslmode=disable` in the client, it will skip the SSLRequest handshake and go straight to plain text, as if `N` had been sent, or TLS had been terminated.

## Original

The first packet from `psql` is `00 00 00 08 04 d2 16 2f`, which I believe is related to StartTLS.

- `openssl s_client -starttls postgres`
- https://www.postgresql.org/docs/current/protocol-message-formats.html
- https://www.tzeejay.com/blog/2022/06/golang-postgresql-check-certificates/

Either way, it doesn't send any SNI or ALPN information until later on in the handshake.

Is there a way that I could route based on the first 8 bytes?

I almost got around this by using `sclient` for TLS and piping the `psql` connection through with `sslmode=disable`, but I think my HTTP matcher is causing it to hang waiting for a `\r\n` that never comes.

mohammed90:
Matching of PostgreSQL connections was recently merged in #186. Anything beyond matching and simple proxying/teeing will require custom handlers or matchers to be developed. My previous research indicates each database has its own packet structure, so each requires its own matching logic:

https://caddy.community/t/need-some-help-with-enabling-lets-encrypt-certificate-on-layer-4-module/18628/11

I can't come up with a UX-friendly way to match arbitrary bytes

coolaj86:
@mohammed90 #186 is not actually detecting Postgres. It's detecting the StartTLS byte (SMTPS, SHTTP).

The "Version 3" byte refers to SSLv3, not pgsql v3.

The `2f` byte refers to SSL/TLS whereas a `30` in that position would refer to GSSAPI (competing standard at the time which postgres also adopted). I'm not sure with all the specifics yet, but it may be a bitmask that signals either protocol.

See: https://github.com/traefik/traefik/issues/9929

coolaj86:
[...]

PR #188, opened 2024-05-04
L4 module for Postgres: matchers

metafeather:
This PR adds additional matchers for the Postgres module:

1. Postgres Matcher can now match on the combination of `user` and `database` parameters when SSL is disabled
2. New PostgresClient Matcher that checks the `application_name` parameter when SSL is disabled
3. New PostgresSSL Matcher that can enforce/reject the use of SSL connections, but cannot match on their content

More tests and docs are in progress, but given the discussion in https://github.com/mholt/caddy-l4/issues/187 I'd like to share the non-SSL matchers, and discuss further enhancements to the PostgresSSL Matcher

### DIFF
--- a/modules/l4postgres/matcher.go
+++ b/modules/l4postgres/matcher.go
@@ -77,8 +77,13 @@ type startupMessage struct {
 	Parameters      map[string]string
 }
 
-// MatchPostgres is able to match Postgres connections.
-type MatchPostgres struct{}
+// MatchPostgres is able to match Postgres connections, optionally further
+// matching on the User or Database being requested
+type MatchPostgres struct {
+	Users     []string
+	Databases []string
+	startup   *startupMessage
+}
 
 // CaddyModule returns the Caddy module information.
 func (MatchPostgres) CaddyModule() caddy.ModuleInfo {
@@ -116,7 +121,7 @@ func (m MatchPostgres) Match(cx *layer4.Connection) (bool, error) {
 	}
 
 	// Try parsing Postgres Params
-	startup := &startupMessage{ProtocolVersion: code, Parameters: make(map[string]string)}
+	m.startup = &startupMessage{ProtocolVersion: code, Parameters: make(map[string]string)}
 	for {
 		k := b.ReadString()
 		if k == "" {
--- a/modules/l4postgres/matcher.go
+++ b/modules/l4postgres/matcher.go
@@ -24,7 +24,7 @@
 //
 //	{
 //		"postgres": {
-//			"users": {
+//			"user": {
 //				"*": ["public_db"],
 //				"alice": ["planets_db", "stars_db"]
 //			}
@@ -32,7 +32,7 @@
 //	}
 //
 //	{
-//		"postgres_clients": ["psql", "TablePlus"]
+//		"postgres_client": ["psql", "TablePlus"]
 //	}
 //
 //	{
@@ -61,7 +61,7 @@ import (
 
 func init() {
 	caddy.RegisterModule(MatchPostgres{})
-	caddy.RegisterModule(MatchPostgresClients{})
+	caddy.RegisterModule(MatchPostgresClient{})
 	caddy.RegisterModule(MatchPostgresSSL{})
 }
 
@@ -89,7 +89,7 @@ func newMessageFromConn(cx *layer4.Connection) (*message, error) {
 
 // MatchPostgres is able to match Postgres connections
 type MatchPostgres struct {
-	Users   map[string][]string
+	User    map[string][]string
 	startup *startupMessage
 }
 
@@ -110,7 +110,7 @@ func (m MatchPostgres) Match(cx *layer4.Connection) (bool, error) {
 	}
 
 	m.startup = newStartupMessage(b)
-	hasConfig := len(m.Users) > 0
+	hasConfig := len(m.User) > 0
 
 	// Finish if this is a SSLRequest and there are no other matchers
 	if m.startup.IsSSL() && !hasConfig {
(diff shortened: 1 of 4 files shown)


mholt:
Thanks for updating this. Will probably give it another review soon.

mholt (commented):
Thanks, this looks quite involved! I'm glad you are doing this :smile: 

I probably won't be testing it myself, so, I imagine you've tested that this works?

One other question I had is: is it possible to have just 1 matcher? I vaguely understand why Postgres SSL is different from regular SSL (but can we call it TLS?). I just wonder if we could simplify the config a bit, or maybe you can help me understand why they should be different matchers entirely... thanks!

sandros94:
> [...] but can we call it TLS? [...]

I wouldn't call it TLS, as it is quite different. postgres (and AFAIK mysql) only support application-level ssl. In simple terms, as I see it, it means that everything should be pre-configured and pre-established.

Looking at some of the pg providers (for example [Neon](https://neon.tech)) they tend to enforce a secured connection forcefully responding `S` on the first request, this makes all client error out if they weren't using a `sslmode` of `required`/`verify-ca`/`verify-full` but making the subsequent request being a TLSv1.3 one with SNI.

Postgres 17 did introduce the ability to directly send the SNI, but before all clients will support that it will require some time.

coolaj86:
You're dealing with sessions, not requests.

If you think of it a requests you logic will not be correct.

HTTP has an abstraction for requests. PG does not.

I'm betting on the future - by the time anything I'm building now is ready for the type of customer who refuses to use pg17 via https://webinstall.dev/psql or whatever other non-apt method, pg17 will be ubiquitous. The customers I'm getting now want other advantages.

tannevaled:
This is a great addition. To tackle the SSL limitation you flagged (matching `user`/`database`/`application_name` is impossible while the connection is encrypted), I put together a small proof-of-concept that complements these matchers rather than duplicating anything.

**The gap:** Postgres doesn't begin TLS with a ClientHello — the client sends the 8-byte `SSLRequest`, waits for `S`/`N`, and only then starts the handshake (as you and others traced in #187). So the generic `tls` handler can't terminate a Postgres connection on its own.

**Proposal — a tiny `postgres_starttls` handler** that does just the Postgres preamble (read `SSLRequest`, reply `S`), then hands off to the existing `tls` handler for the actual termination. It deliberately reuses `l4tls` (no TLS code copied). After that, your matchers here match on the cleartext startup message:

```
route {
    postgres
    postgres_starttls   # read SSLRequest, reply 'S'
    tls { ... }         # existing l4tls terminator
    postgres {          # this PR's matchers, now on cleartext
        database acme
    }
    proxy ...
}
```

PoC branch (handler + tests, builds/tests/lint clean): https://github.com/tannevaled/caddy-l4/tree/feat/postgres-starttls

Two notes:
- This is **terminate-only**. Re-originating SSL to an upstream that expects it (the general case) needs the same `SSLRequest`→`S` dance *outbound*, which the proxy's plain upstream-TLS doesn't do — I'd keep that as a separate, larger follow-up. For deployments where the hop to the backend is already on a trusted/encrypted network, terminate-only is enough.
- It pairs naturally with your `PostgresSSL` matcher to gate on "is an SSLRequest" before terminating.

Happy to turn it into a proper PR (stacked on this one, so its matchers become usable behind SSL), or adjust naming/placement — whatever fits best. @mholt @francislavoie does this direction look reasonable?

tannevaled:
Follow-up: I also pushed the **outbound** counterpart to the same branch, so the full picture is testable end-to-end:

- inbound: `postgres_starttls` handler → `tls` terminates → your matchers see cleartext;
- outbound: `StartTLSClient(conn, cfg)` re-originates SSL to a Postgres upstream that expects `sslmode != disable` (sends `SSLRequest`, expects `S`, then a `tls.Client` handshake).

I deliberately kept the outbound part a **self-contained function rather than wiring it into `l4proxy`**, to avoid coupling the generic proxy to Postgres. It would slot in behind an opt-in upstream STARTTLS hook — open to ideas on the cleanest shape (a generic `starttls <proto>` upstream option vs. a postgres-specific one). Terminate-only setups (backend reached over a trusted/encrypted network, e.g. a WireGuard overlay) don't need the outbound side at all.

Branch (both directions, tests + handshake test, green): https://github.com/tannevaled/caddy-l4/tree/feat/postgres-starttls

vnxme:
@tannevaled So shall we still consider this PR as a candidate for merging, or #432 effectively replaces this PR?

tannevaled:
@vnxme #188 should stay — the two PRs are complementary, not alternatives. #188 adds the matchers (user/database/application_name), and #432 adds the handler that decrypts STARTTLS traffic so those matchers actually fire on TLS-encrypted Postgres connections (the SSL limitation you and the prior threads flagged). Without #188, #432 only opens the cleartext window; without #432, #188's matchers don't reach TLS-protected sessions. Sequencing-wise I'd suggest landing #188 first (it stands on its own for unencrypted connections) and then #432 on top, but either order works since they don't overlap in code. Happy to rebase #188 against current master if that helps the review.

francislavoie:
I think it's fine if you want to open a new PR to replace #188, just make sure @metafeather gets the author/co-author credit.

tannevaled:
Thanks @francislavoie — that works. Plan: I'll open a fresh matchers PR that supersedes #188 (rebased on current master, with @metafeather kept as `Co-authored-by:` on the commits so the original authorship is preserved in git history), and keep #432 (the STARTTLS handler) as a separate PR on top. Will close #188 when the new one lands. I'll ping you and @vnxme once the matchers PR is up.

tannevaled:
As promised, the matchers PR is up: #436. It supersedes this one, rebased onto the current hardened `MatchPostgres` on `master`, with you (@metafeather) kept as `Co-authored-by:` on the commit so the authorship is preserved in git history.

It keeps the same three matchers, module IDs and JSON config shapes as here (existing configs keep working), and additionally:
- reuses master's DoS-hardened read for all three matchers;
- adds `json` tags so the lowercase keys (`user`, `client`, `disabled`) are produced;
- adds `UnmarshalCaddyfile` (this PR was JSON-only) with `caddyfile_adapt` tests;
- brings the package to 100% coverage (the existing `TestMatchPostgres` is unchanged and still green) and adds docs.

The `postgres_starttls` handler stays as its own PR (#432) on top. I'll close this one once #436 lands. @francislavoie @vnxme — review welcome whenever it suits you; no rush.

PR #192, opened 2024-05-06
Non blocking matchers & matching timeout

ydylla:
This is a better version of #72 that actually works 😄   
Instead of the wrapped route chain routing/matching is now done in a loop until a timeout is reached. Each connection is prefetched without blocking. Then all routes and their matchers are tried on the available data. If nothing matched more data is prefetched and all routes are tried again. This is done until the timeout is reached. The timeout can be configured per layer4 server and defaults to 3s. I first had it set to 500ms but since the matching phase may contain tls handshakes that was too short. I saw some failed connections from far away servers where tls handshaking apparently takes about 1000ms.  

If you want to test or compare both prefetching branches you can use this caddy config:

<details>
<summary>prefetch.json</summary>

```json
{
  "admin": {
    "disabled": true
  },
  "logging": {
    "logs": {
    }
  },
  "apps": {
    "tls": {
      "certificates": {
        "automate": ["localhost"]
      },
      "automation": {
        "policies": [{
          "subjects": ["localhost"],
          "issuers": [{
            "module": "internal"
          }]
        }]
      }
    },
    "layer4": {
      "servers": {
        "https": {
          "listen": ["0.0.0.0:10443"],
          "routes": [
            {
              "match": [
                {"proxy_protocol": {}}
              ],
              "handle": [
                {
                  "handler": "proxy_protocol",
                  "timeout": "2s",
                  "allow": ["127.0.0.1/32"]
                }
              ]
            },
            {
              "match": [
              ],
              "handle": [
                {"handler": "tls"}
              ]
            },
            {
              "match": [
                {"http": [{"host": ["localhost"]}]}
              ],
              "handle": [
                {
                  "handler": "proxy",
                }
              ]
            }
          ]
        }
      }
    },
    "http": {
      "servers": {
        "backend": {
          "routes": [
            {
              "handle": [{
                "handler": "static_response",
                "status_code": "200",
                "body": "Hello World\n",
                "headers": {
                }
              }]
            }
          ]
        }
      }
    }
  }
}
```
</details>

The following curl commands should all work with this branch while some do not work with the old version:

```shell
curl -v -k --http2 https://localhost:10443
curl -v -k --http2-prior-knowledge https://localhost:10443
curl -v -k http://localhost:10443
curl -v -k --haproxy-protocol https://localhost:10443
```

I hope this is not a too big rewrite and you are still interested in merging it. I currently also test it with real browser traffic and until now it works fine.

### DIFF
--- a/layer4/connection.go
+++ b/layer4/connection.go
@@ -138,13 +138,13 @@ func (cx *Connection) prefetch() (err error) {
 
 	for len(cx.buf) < MaxMatchingBytes {
 		free := cap(cx.buf) - len(cx.buf)
-		if free >= PrefetchChunkSize {
-			n, err = cx.Conn.Read(cx.buf[len(cx.buf) : len(cx.buf)+PrefetchChunkSize])
+		if free >= prefetchChunkSize {
+			n, err = cx.Conn.Read(cx.buf[len(cx.buf) : len(cx.buf)+prefetchChunkSize])
 			cx.buf = cx.buf[:len(cx.buf)+n]
 		} else {
 			if tmp == nil {
 				tmp = bufPool.Get().([]byte)
-				tmp = tmp[:PrefetchChunkSize]
+				tmp = tmp[:prefetchChunkSize]
 				defer bufPool.Put(tmp)
 			}
 			n, err = cx.Conn.Read(tmp)
@@ -157,7 +157,7 @@ func (cx *Connection) prefetch() (err error) {
 			return err
 		}
 
-		if n < PrefetchChunkSize {
+		if n < prefetchChunkSize {
 			break
 		}
 	}
@@ -212,7 +212,7 @@ func (cx *Connection) GetVar(key string) interface{} {
 }
 
 // MatchingBytes returns all bytes currently available for matching. This is only intended for reading.
-// Do not write into the slice it's a view of the internal buffer and you will likely mess up the connection.
+// Do not write into the slice. It's a view of the internal buffer and you will likely mess up the connection.
 func (cx *Connection) MatchingBytes() []byte {
 	return cx.buf[cx.offset:]
 }
@@ -229,7 +229,7 @@ var (
 	listenerCtxKey caddy.CtxKey = "listener"
 )
 
-const PrefetchChunkSize = 1024
+const prefetchChunkSize = 1024
 
 // MaxMatchingBytes is the amount of bytes that are at most prefetched during matching.
 // This is probably most relevant for the http matcher since http requests do not have a size limit.
@@ -238,6 +238,6 @@ const MaxMatchingBytes = 8 * 1024
 
 var bufPool = sync.Pool{
 	New: func() interface{} {
-		return make([]byte, 0, PrefetchChunkSize)
+		return make([]byte, 0, prefetchChunkSize)
 	},
 }
--- a/layer4/connection.go
+++ b/layer4/connection.go
@@ -213,8 +213,8 @@ func (cx *Connection) GetVar(key string) interface{} {
 
 // MatchingBytes returns all bytes currently available for matching. This is only intended for reading.
 // Do not write into the slice it's a view of the internal buffer and you will likely mess up the connection.
-func (cx *Connection) MatchingBytes() ([]byte, error) {
-	return cx.buf[cx.offset:], nil
+func (cx *Connection) MatchingBytes() []byte {
+	return cx.buf[cx.offset:]
 }
 
 var (
--- a/layer4/connection.go
+++ b/layer4/connection.go
@@ -65,9 +65,10 @@ type Connection struct {
 
 	Logger *zap.Logger
 
-	buf      []byte // stores matching data
-	offset   int
-	matching bool
+	buf          []byte // stores matching data
+	offset       int
+	frozenOffset int
+	matching     bool
 
 	bytesRead, bytesWritten uint64
 }
@@ -80,18 +81,22 @@ var ErrMatchingBufferFull = errors.New("matching buffer is full")
 // and once depleted (or if there isn't one), it continues
 // reading from the underlying connection.
 func (cx *Connection) Read(p []byte) (n int, err error) {
-	if cx.matching {
-		if len(cx.buf) == 0 || len(cx.buf[cx.offset:]) == 0 {
-			return 0, ErrConsumedAllPrefetchedBytes
-		}
+	// if we are matching and consumed the buffer exit with error
+	if cx.matching && (len(cx.buf) == 0 || len(cx.buf) == cx.offset) {
+		return 0, ErrConsumedAllPrefetchedBytes
 	}
 
 	// if there is a buffer we should read from, start
 	// with that; we only read from the underlying conn
 	// after the buffer has been "depleted"
-	if cx.offset < len(cx.buf) {
+	if len(cx.buf) > 0 && cx.offset < len(cx.buf) {
 		n := copy(p, cx.buf[cx.offset:])
 		cx.offset += n
+		if !cx.matching && cx.offset == len(cx.buf) {
+			// if we are not in matching mode reset buf automatically after it was consumed
+			cx.offset = 0
+			cx.buf = cx.buf[:0]
+		}
 		return n, nil
 	}
 
@@ -174,21 +179,14 @@ func (cx *Connection) prefetch() (err error) {
 // freeze activates the matching mode that only reads from cx.buf.
 func (cx *Connection) freeze() {
 	cx.matching = true
+	cx.frozenOffset = cx.offset
 }
 
 // unfreeze stops the matching mode and resets the buffer offset
 // so that the next reads come from the buffer first.
 func (cx *Connection) unfreeze() {
 	cx.matching = false
-	cx.offset = 0
-}
-
-// clear discards cx.buf and resets cx.offset to prepare the connection for the next prefetch & matching phase
-func (cx *Connection) clear() {
-	if cx.buf != nil {
-		cx.buf = cx.buf[:0]
-	}
-	cx.offset = 0
+	cx.offset = cx.frozenOffset
 }
 
 // SetVar sets a value in the context's variable table with
--- a/layer4/connection.go
+++ b/layer4/connection.go
@@ -15,9 +15,8 @@
 package layer4
 
 import (
-	"bytes"
 	"context"
-	"io"
+	"errors"
 	"net"
 	"sync"
 
@@ -30,7 +29,7 @@ import (
 // and variable table. This function is intended for use at the start of a
 // connection handler chain where the underlying connection is not yet a layer4
 // Connection value.
-func WrapConnection(underlying net.Conn, buf *bytes.Buffer, logger *zap.Logger) *Connection {
+func WrapConnection(underlying net.Conn, buf []byte, logger *zap.Logger) *Connection {
 	repl := caddy.NewReplacer()
 	repl.Set("l4.conn.remote_addr", underlying.RemoteAddr())
 	repl.Set("l4.conn.local_addr", underlying.LocalAddr())
@@ -66,50 +65,40 @@ type Connection struct {
 
 	Logger *zap.Logger
 
-	buf       *bytes.Buffer // stores recordings
-	bufReader io.Reader     // used to read buf so it doesn't discard bytes
-	recording bool
+	buf      []byte // stores matching data
+	offset   int
+	matching bool
 
 	bytesRead, bytesWritten uint64
 }
+
+var ErrConsumedAllPrefetchedBytes = errors.New("consumed all prefetched bytes")
+var ErrMatchingBufferFull = errors.New("matching buffer is full")
 
 // Read implements io.Reader in such a way that reads first
 // deplete any associated buffer from the prior recording,
 // and once depleted (or if there isn't one), it continues
 // reading from the underlying connection.
 func (cx *Connection) Read(p []byte) (n int, err error) {
+	if cx.matching {
+		if len(cx.buf) == 0 || len(cx.buf[cx.offset:]) == 0 {
+			return 0, ErrConsumedAllPrefetchedBytes
+		}
+	}
+
 	// if there is a buffer we should read from, start
 	// with that; we only read from the underlying conn
 	// after the buffer has been "depleted"
-	if cx.bufReader != nil {
-		n, err = cx.bufReader.Read(p)
-		if err == io.EOF {
-			cx.bufReader = nil
-			err = nil
-		}
-		// prevent first read from returning 0 bytes because of empty bufReader
-		if !(n == 0 && err == nil) {
-			return
-		}
+	if cx.offset < len(cx.buf) {
+		n := copy(p, cx.buf[cx.offset:])
+		cx.offset += n
+		return n, nil
 	}
 
 	// buffer has been "depleted" so read from
 	// underlying connection
 	n, err = cx.Conn.Read(p)
 	cx.bytesRead += uint64(n)
-
-	if !cx.recording {
-		return
-	}
-
-	// since we're recording at this point, anything that
-	// was read needs to be written to the buffer, even
-	// if there was an error
-	if n > 0 {
-		if nw, errw := cx.buf.Write(p[:n]); errw != nil {
-			return nw, errw
-		}
-	}
 
 	return
 }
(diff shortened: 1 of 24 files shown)


layer4/routes.go
```diff
@@ -92,75 +104,70 @@ func (routes RouteList) Provision(ctx caddy.Context) error {
...
+						handler = route.middleware[i](handler)
+					}
+					err = handler.Handle(cx)
+					if err != nil {
+						return err
+					}
+
+					// If matched route is terminal we stop routing,
+					// otherwise we jump back to the start of the routing loop to peel of more protocol layers.
+					if route.Terminal {
```
mholt, line 164: Is there a way to do this without explicitly hard-coding whether a handler is terminal?

One problem I found when designing Caddy's HTTP routes and handlers was that, at first, I thought a handler was either terminal or not. For example, a file server was terminal because it loads the file from disk and writes the response. Nothing can come after it up the chain. An encoder module was never terminal, as it always encoded whatever the terminal handler up the chain would write.

But a cache handler was sometimes terminal and sometimes not. If it was a cache hit, it was terminal; if a miss, it would NOT be terminal because it would have to send the connection up the chain to be handled by the handler originating the content.

So one concern I have with this design is that it requires a handler to either be terminal or not, always.

Is there a way we can remove that `IsTerminal()` method and let it be determined implicitly by the flow of handling whether a handler is terminal? Even if it means putting a handler at the end that is a no-op and is always terminal.

modules/l4subroute/handler.go
```diff
@@ -34,6 +35,9 @@ type Handler struct {
 	// The primary list of routes to compile and execute.
 	Routes layer4.RouteList `json:"routes,omitempty"`
 
+	// Maximum time connections have to complete the matching phase (the first terminal handler is matched). Default: 3s.
+	MatchingTimeout caddy.Duration `json:"matching_timeout,omitempty"`
```
mholt, line 39: Would this be, like, 3s + 3s for the main route timeout + subroute? Or does the shorter of the two "win"?

mholt (commented):
Ok, I might do this in chunks/iterations. Starting with the first things I notice, these are just questions I have -- overall I think this is a very thoughtful change so it's looking quite good. I just want to make sure I understand it and that we eek out the best possible design. :100:

layer4/routes.go
ydylla: Yes I was also not a fan of IsTerminal. Now that you said it I think we can indeed get rid of it again. Because we already have a handler to catch wrapped connection (see [lastHandler](https://github.com/mholt/caddy-l4/pull/192/files#diff-ca91bc2c00b6e9985b0eeab47b53d0d3742328972c06b743ef18842f0f635cffR144)). I am sure we can also use it as terminal marker. Thanks for the idea I will try it tomorrow.

modules/l4subroute/handler.go
ydylla: After the subroute handler is matched the deadline from the root handler is removed. Subroute will then start its own matching timeout. So yes it could be 3s + 3s. But if the subroute is not terminal we jump back to the main routing loop which resets the timeout to the old initial deadline. Which means it may immediately time out.

mholt: oh ok, so if it spends its time matching in the subroute, it won't go back and add 3 more seconds in the main route. That's good then :+1:

layer4/routes.go
ydylla: @mholt I removed the IsTerminal function again. If a route is terminal or not is now determined by if the `lastHanlder` was called or not.

modules/l4subroute/handler.go
```diff
@@ -49,6 +53,10 @@ func (Handler) CaddyModule() caddy.ModuleInfo {
 func (h *Handler) Provision(ctx caddy.Context) error {
 	h.logger = ctx.Logger(h)
 
+	if h.MatchingTimeout <= 0 {
+		h.MatchingTimeout = caddy.Duration(3 * time.Second)
```
mholt, line 57: Let's make the `3 * time.Second` a const since I think we use it 3 times

layer4/routes.go
```diff
@@ -92,75 +98,71 @@ func (routes RouteList) Provision(ctx caddy.Context) error {
...
-			// if we don't make this copy, we'd affect the underlying
-			// pointer for all future request (yikes); we could
-			// alternatively solve this by moving the func below out of
-			// this closure and into a standalone package-level func,
-			// but I just thought this made more sense
-			nextCopy := next
-
-			// route must match at least one of the matcher sets
-			matched, err := route.matcherSets.AnyMatch(cx)
+func (routes RouteList) Compile(next NextHandler, logger *zap.Logger, matchingTimeout time.Duration) Handler {
```
mholt, line 101: I wonder if the aesthetics of calling this would be improved a bit if `next` is the last argument. That way we don't have a potential multi-line, inlined handler method that is followed by 2 more arguments after it.

mholt (commented):
Ok did another quick scan -- looking better! I love that we don't have `IsTerminal()` anymore. I'm glad to see you were able to make that work :smiley:

I still need to fully grok the "freeze/unfreeze" stuff (as opposed to recording?) but I'll do another pass soon :) Thank you!

layer4/routes.go
mholt: Awesomeeee

ydylla:
> I still need to fully grok the "freeze/unfreeze" stuff (as opposed to recording?)

I renamed them because "record" felt wrong since there is no recording happening during matching anymore. Instead the connection is frozen and can only read data from the prefetched buffer. Technically a matcher could still write but that would very likely result in an invalid state. We could return an error if writes are attempted during matching to prevent misuse.

layer4/connection.go
```diff
@@ -187,8 +229,15 @@ var (
 	listenerCtxKey caddy.CtxKey = "listener"
 )
 
+const PrefetchChunkSize = 1024
```
mholt, line 232: Does this need to be exported?

Should this be configurable instead?

layer4/connection.go
```diff
@@ -167,14 +203,20 @@ func (cx Connection) SetVar(key string, value interface{}) {
...
+func (cx *Connection) GetVar(key string) interface{} {
 	varMap, ok := cx.Context.Value(VarsCtxKey).(map[string]interface{})
 	if !ok {
 		return nil
 	}
 	return varMap[key]
 }
 
+// MatchingBytes returns all bytes currently available for matching. This is only intended for reading.
+// Do not write into the slice it's a view of the internal buffer and you will likely mess up the connection.
```
mholt, line 215:
```suggestion
// Do not write into the slice. It's a view of the internal buffer and you will likely mess up the connection.
```

mholt (commented):
That makes sense. Thanks for the explanation!

Ok I think we're close. I understand things a little better. Still a few more questions but nothing too major. Almost there!! Thanks for hanging in there with me!

modules/l4http/httpmatcher.go
```diff
@@ -78,11 +80,18 @@ func (m MatchHTTP) Match(cx *layer4.Connection) (bool, error) {
 	req, ok := cx.GetVar("http_request").(*http.Request)
 	if !ok {
 		var err error
-		bufReader := bufio.NewReader(cx)
+
+		data := cx.MatchingBytes()
+		if !m.isHttp(data) {
```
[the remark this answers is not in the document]
ydylla: No it also matches http2 (see [TestMatchHTTP_isHttp](https://github.com/mholt/caddy-l4/pull/192/files#diff-0f169956149b8041f34a0b167acc8d7cdcbe68329fdb36cfd333c9f403ae7e79R273)), since http2 requests start http1 parser friendly/compatible.

layer4/connection.go
ydylla: No its an implementation detail. And also just an attempt to keep the buffers smaller, since most matching will not need the full 8 KiB max buffer size. If we make something configurable then the max matching buffer size, but for now I do not think this is necessary.

mholt: Ok cool

mholt (approved):
I think this is looking good :smiley: Are you able to help with it if anyone reports an issue related to this? I'll be happy to merge this and we can try it out! Excellent work.

ydylla:
Yes I can help bug fixing it, especially as long as I use it myself 😄

mholt:
Awesome, thanks!!

T4cC0re:
This PR breaks our configuration.  
A TLS connection is never established successfully with the module on this PR and Caddy 2.7.6. It works fine on c5d815daed3cf7c9be6f199b324aeae8b3c865f0 with Caddy 2.7.6.

We use caddy-l4 to forward to a Unix socket depending on the TLS Handshake, then have a secondary listened on a UNIX socket with on-demand TLS provisioning. (We are doing it this way , to be able to use different TLS provisioning policies, and others)

Setting the new timeout parameters to something like '15m' does not help.

The config:

```json
{
  "logging": {
    "logs": {
      "default": {
        "level": "info",
        "writer": {
          "output": "stdout"
        },
        "encoder": {
          "format": "json"
        }
      }
    }
  },
  "apps": {
    "layer4": {
      "servers": {
        "http": {
          "listen": [
            ":80"
          ],
          "routes": [
            {
              "match": [
                {
                  "http": [
                    {
                      "host": [
                        "gitlab.REDACTED"
                      ]
                    }
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "proxy",
                  "proxy_protocol": "v2",
                  "upstreams": [
                    {
                      "dial": [
                        "unix//var/opt/caddy/caddy.gitlab"
                      ]
                    }
                  ]
                }
              ]
            },
            {
              "match": [
                {
                  "http": [
                    {
                      "host": [
                        "registry.REDACTED"
                      ]
                    }
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "proxy",
                  "proxy_protocol": "v2",
                  "upstreams": [
                    {
                      "dial": [
                        "unix//var/opt/caddy/caddy.registry"
                      ]
                    }
                  ]
                }
              ]
            }
          ]
        },
        "https": {
          "listen": [
            ":443"
          ],
          "routes": [
            {
              "match": [
                {
                  "tls": {}
                }
              ],
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "match": [
                        {
                          "tls": {
                            "sni": [
                              "gitlab.REDACTED"
                            ]
                          }
                        }
                      ],
                      "handle": [
                        {
                          "handler": "proxy",
                          "proxy_protocol": "v2",
                          "upstreams": [
                            {
                              "dial": [
                                "unix//var/opt/caddy/caddy.gitlab"
                              ]
                            }
                          ]
                        }
                      ]
                    },
                    {
                      "match": [
                        {
                          "tls": {
                            "sni": [
                              "registry.REDACTED"
                            ]
                          }
                        }
                      ],
                      "handle": [
                        {
                          "handler": "proxy",
                          "proxy_protocol": "v2",
                          "upstreams": [
                            {
                              "dial": [
                                "unix//var/opt/caddy/caddy.registry"
                              ]
                            }
                          ]
                        }
                      ]
                    }
                  ]
                }
              ]
            }
          ]
        },
        "ssh": {
          "listen": [
            ":22"
          ],
          "routes": [
            {
              "handle": [
                {
                  "handler": "proxy",
                  "proxy_protocol": "v2",
                  "upstreams": [
                    {
                      "dial": [
                        "10.0.1.1:2022"
                      ]
                    }
                  ]
                }
              ]
            }
          ]
        }
      }
    },
    "http": {
      "http_port": 80,
      "https_port": 443,
      "servers": {
        "gitlab": {
          "strict_sni_host": true,
          "listen": [
            "unix//var/opt/caddy/caddy.gitlab"
          ],
          "listener_wrappers": [
            {
              "wrapper": "proxy_protocol",
              "timeout": 0
            },
            {
              "wrapper": "http_redirect"
            },
            {
              "wrapper": "tls"
            }
          ],
          "metrics": {},
          "routes": [
            {
              "match": [
                {
                  "host": [
                    "gitlab.REDACTED"
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "headers",
                  "response": {
                    "set": {
                      "Server": [
                        "REDACTED"
                      ]
                    }
                  }
                },
                {
                  "encodings": {
                    "gzip": {},
                    "zstd": {}
                  },
                  "handler": "encode",
                  "prefer": [
                    "zstd",
                    "gzip"
                  ]
                },
                {
                  "handler": "reverse_proxy",
                  "load_balancing": {
                    "selection_policy": {
                      "policy": "round_robin"
                    }
                  },
                  "upstreams": [
                    {
                      "dial": "10.0.1.1:8081"
                    }
                  ]
                }
              ],
              "terminal": true
            }
          ]
        },
        "registry": {
          "strict_sni_host": true,
          "listen": [
            "unix//var/opt/caddy/caddy.registry"
          ],
          "listener_wrappers": [
            {
              "wrapper": "proxy_protocol",
              "timeout": 0
            },
            {
              "wrapper": "http_redirect"
            },
            {
              "wrapper": "tls"
            }
          ],
          "metrics": {},
          "routes": [
            {
              "match": [
                {
                  "host": [
                    "registry.REDACTED"
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "headers",
                  "response": {
                    "set": {
                      "Server": [
                        "REDACTED"
                      ]
                    }
                  }
                },
                {
                  "encodings": {
                    "gzip": {},
                    "zstd": {}
                  },
                  "handler": "encode",
                  "prefer": [
                    "zstd",
                    "gzip"
                  ]
                },
                {
                  "handler": "reverse_proxy",
                  "load_balancing": {
                    "selection_policy": {
                      "policy": "round_robin"
                    }
                  },
                  "upstreams": [
                    {
                      "dial": "10.0.1.1:18081"
                    }
                  ]
                }
              ],
              "terminal": true
            }
          ]
        }
      }
    },
    "tls": {
      "certificates": {
        "load_folders": [
          "/var/opt/caddy/certificates"
        ]
      },
      "automation": {
        "policies": [
          {
            "on_demand": true,
            "issuers": [
              {
                "email": "REDACTED",
                "module": "acme"
              },
              {
                "email": "REDACTED",
                "module": "zerossl"
              }
            ]
          }
        ],
        "on_demand": {
          "ask": "http://127.0.0.1:2223"
        }
      }
    }
  }
}
```

mholt:
@T4cC0re Is there any particular error message? "A TLS connection is never established successfully" is not a lot to go on...

ydylla:
I think I know what's happening. The subroute is blocking on its prefetch call because the root route already fetched all bytes from the client. The compiled route in the subroute handler should probably not call prefetch if there is already some data available (at least not for the first matcher loop iteration). Sorry I overlooked this, I never actually tested the subroute handler.

@T4cC0re You can work around this by not using the subroute handler. You should be able to just copy your subroute "routes" to the root "routes".

ydylla:
> The subroute is blocking on its prefetch call because the root route already fetched all bytes from the client.

This is exactly what happens. https://github.com/mholt/caddy-l4/pull/194 fixes it.

T4cC0re:
> This is exactly what happens. #194 fixes it.

Can confirm. It works now :) Thanks for the quick turnaround!

mholt:
Really appreciate your help @ydylla -- thank you!

mholt:
@ydylla One more potential bug with this commit: have you considered the case where a TLS connection is wrapped? i.e.:

```json
{
	"apps": {
		"layer4": {
			"servers": {
				"signal_proxy": {
					"listen": [
						":443"
					],
					"routes": [
						{
							"handle": [
								{
									"handler": "tls"
								}
							]
						},
						{
							"match": [
								{
									"tls": {
										"sni": [
											"chat.signal.org",
											"ud-chat.signal.org"
										]
									}
								}
							],
							"handle": [
								{
									"handler": "proxy",
									"upstreams": [
										{
											"dial": [
												"chat.signal.org:443"
											]
										}
									]
								}
							]
						},
						{
							"match": [
								{
									"tls": {
										"sni": [
											"storage.signal.org",
											"cdn.signal.org",
											"cdn2.signal.org",
											"cdn3.signal.org",
											"cdsi.signal.org",
											"contentproxy.signal.org",
											"uptime.signal.org",
											"sfu.voip.signal.org",
											"svr2.signal.org",
											"updates.signal.org",
											"updates2.signal.org",
											"backend1.svr3.signal.org",
											"backend2.svr3.signal.org",
											"backend3.svr3.signal.org"
										]
									}
								}
							],
							"handle": [
								{
									"handler": "proxy",
									"upstreams": [
										{
											"dial": [
												"{l4.tls.server_name}:443"
											]
										}
									]
								}
							]
						}
					]
				}
			}
		},
		"tls": {
			"certificates": {
				"automate": [
					"test-1.finerpixels.com"
				]
			}
		}
	}
}
```

A Signal proxy first terminates the "outer" layer of TLS, then with that encrypted tunnel, it sends an inner TLS connection. This config used to be able to match the ServerName (SNI) of the _inner_ TLS connection without terminating it, but I noticed that with this change, it seems to try to terminate the inner connection.

Do you think you could help fix this? It appears to be a regression. (Sorry!)

ydylla:
@mholt I don't think the wrapping is the problem. It's likely my rather arbitrary choice of jumping back to the first route after a match instead of continuing. It was easier to code and seamed reasonable. If you add an "example.com" sni matcher to the first route with the tls handler it should work.  
I will try to reproduce and fix it, should be possible. And don't worry I also want the stuff I write to work 😄

mholt:
Oh you're good! I super appreciate your contributions (I've been stretched a bit thin lately.) Let me know if I can help at all. Thanks so much :sweat_smile:

ydylla:
@mholt #196 recreates the old matching behavior.  
I spend the most time trying to come up with a command line that does TLS in TLS with curl, echo and openssl s_client but eventually I gave up. Turns out writing it in go was way easier 😅 . In case anyone also needs this:
```go
package main

import (
	"crypto/tls"
	"log"
)

func main() {
	outerConn, err := tls.Dial("tcp", "127.0.0.1:10443", &tls.Config{
		ServerName:         "outer",
		InsecureSkipVerify: false,
	})
	if err != nil {
		log.Fatal(err)
	}

	innerConn := tls.Client(outerConn, &tls.Config{
		ServerName:         "inner",
		InsecureSkipVerify: false,
	})

	_, err = innerConn.Write([]byte("GET / HTTP/1.0\r\nHost: localhost\r\n\r\n"))
	if err != nil {
		log.Fatal(err)
	}
	buf := make([]byte, 1024)
	n, err := innerConn.Read(buf)
	if err != nil {
		log.Fatal(err)
	}
	println(string(buf[:n]))
}
```

mholt:
Ah :sweat_smile: Thanks for working on that and posting your test program. I will review the patch soon!! Thank you!

PR #194, opened 2024-05-25
fix: subroute blocking if all client data was already read

ydylla:
See https://github.com/mholt/caddy-l4/pull/192#issuecomment-2129307974


mholt (approved):
This is great, thanks for the fix, and for the helpful explanatory comments!

PR #196, opened 2024-06-02
fix: skip already seen routes after a match

ydylla:
This is done to be backwards compatible with the old matching behavior see https://github.com/mholt/caddy-l4/pull/192#issuecomment-2143681952


ydylla:
Sorry for the force pushs. I added
```Go
// Only skip once after a match, so it behaves like we continued after the match.
lastMatchedRouteIdx = -1
```
and moved the `lastMatchedRouteIdx = i` line to a better place.  
It gets increasingly difficult to keep all situations in mind for this function.

mholt (approved):
Thanks for the patch. Indeed, routes should be invoked sequentially, in step with their matchers; so matching doesn't restart the chain.

Hopefully that's as complex as it has to get! :sweat_smile:

PR #206, opened 2024-06-26
handlers(l4proxy): add use_client_config to l4proxy upstream conf

legobeat:
The TLS client configuration for `proxy` `upstreams` considers the downstream connectino values only if configured empty.

This adds a new configuration value `use_client_conf`, which allows preserving this behavior despite configuring the `tls` field.

### DIFF
--- a/modules/l4proxy/proxy.go
+++ b/modules/l4proxy/proxy.go
@@ -201,7 +201,7 @@ func (h *Handler) dialPeers(upstream *Upstream, repl *caddy.Replacer, down *laye
 			// in which case we adopt the downstream client's TLS ClientHello for ours;
 			// i.e. by default, make the client's TLS config as transparent as possible
 			tlsCfg := upstream.tlsConfig
-			if tlsCfg == nil {
+			if tlsCfg == nil || upstream.TLS.PropagateClientConfig {
 				tlsCfg = new(tls.Config)
 				if hellos := l4tls.GetClientHelloInfos(down); len(hellos) > 0 {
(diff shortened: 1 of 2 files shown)


mholt:
Hmm, thanks -- I'm trying to understand when this would be useful though? If the server has a TLS client configured, then it is used, otherwise we transparently use the client's TLS config. How can we possibly use both?

mholt:
> For example, caddy-l4 user may want to override renegotiation or alpn/protocol options but still otherwise fill in from the client? 

I think I see, so you want to use the client's connection properties but just change one or two of them.

If so, then yeah, the possibility of empty and nil values colliding will be tricky without some annoying changes... hmm.

vnxme:
@legobeat thank you for contributing!

Actually it was a bug that the TLS upstreams copied the downstream TLS config only if no options other than `tls` were set - it's fixed with #393. Now we also copy the downstream TLS config if any `tls_*` option is configured for the upstream. But server names shouldn't be adopted from the downstream unless explicitly set with `tls_server_name` - this is what #398 deals with.

You also mentioned changing application-layer protocols for the upstream. This is a limitation we have in Caddy [here](https://github.com/caddyserver/caddy/blob/8499e34e10c4f5d8445d8d3cca3338cbb13bc2e1/modules/caddyhttp/reverseproxy/httptransport.go#L629-L690), but we can implement it as well if there is much demand. PRs are welcome: we would need one on Caddy and another one here.

Closing this PR as not planned.

PR #216, opened 2024-07-09
Optimized readability

Minecodes:
I ran the examples through Prettier in VSCode to optimize their readability because the tabs were too wide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,22 +80,20 @@ A simple echo server:
 
 ```json
 {
-	"apps": {
-		"layer4": {
-			"servers": {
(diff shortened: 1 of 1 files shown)


mholt:
Isn't tab width a viewer setting? For example in Firefox you can change your tab width to a lower value. I'm sure Chrome and other browsers probably have similar settings.

I'm not a huge fan of forcing a certain width on everyone by using spaces.

mholt:
Going to close this since I don't see a compelling argument for why this needs to be merged. :man_shrugging:

PR #218, opened 2024-07-18
Rename ip handler into remote_ip

vnxme:
As discussed in mholt/caddy-l4#217, this is a breaking change that renames `ip` handler into `remote_ip`. 

Consequently, we will have consistently named  `remote_ip` matchers at layer4, tls and http levels.

If this PR gets merged first, I will rebase mholt/caddy-l4#217 on it. Otherwise this PR will have to be rebased.


mholt (approved):
Looks good. Thanks for bringing this to my attention!

@francislavoie  what do you think?

francislavoie:
Seems fine to me. One minor concern is we don't really have release notes or changelogs so it might be a surprise for some to have to fix it. But I agree with doing it now for consistency.

PR #220, opened 2024-07-22
Regular expression (regexp) matcher

vnxme:
### Summary

The idea behind `regexp` matcher is to allow low-level packet matching. It's pretty simple under the hood, but actually very powerful. With this matcher Caddy can multiplex virtually any protocol, if a regular expression syntax is enough to distinguish the desired one from others.

### Code examples

Proxy a connection starting with a 0xFFEEDD000000 sequence to another machine
```json
{
  "apps": {
    "layer4": {
      "servers": {
        "example": {
          "listen": ["0.0.0.0:443"],
          "routes": [
            {
              "handle": [
                {
                  "handler": "proxy",
                  "upstreams": [
                    {"dial": ["another.machine.local:456"]}
                  ]
                }
              ],
              "match": [
                {"regexp": {"count": 6, "pattern": "^\\xFF\\xEE\\xDD\\x00\\x00\\x00$"}}
              ]
            },
            {
              "handle": [
                {
                  "handler": "tls"
                },
                {
                  "handler": "proxy",
                  "upstreams": [
                    {"dial": ["localhost:8080"]}
                  ]
                }
              ],
              "match": [
                {"tls": {}}
              ]
            }
          ]
        }
      }
    }
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ Current matchers:
 - **layer4.matchers.not** - matches connections that aren't matched by inner matcher sets.
 - **layer4.matchers.proxy_protocol** - matches connections that start with [HAPROXY proxy protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt).
 - **layer4.matchers.rdp** - matches connections that look like [RDP](https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-RDPBCGR/%5BMS-RDPBCGR%5D.pdf).
+- **layer4.matchers.regexp** - matches connections that have the first packet bytes matching a regular expression.
 - **layer4.matchers.socks4** - matches connections that look like [SOCKSv4](https://www.openssh.com/txt/socks4.protocol).
 - **layer4.matchers.socks5** - matches connections that look like [SOCKSv5](https://www.rfc-editor.org/rfc/rfc1928.html).
 - **layer4.matchers.xmpp** - matches connections that look like [XMPP](https://xmpp.org/about/technology-overview/).
--- a/modules/l4regexp/matcher.go
+++ b/modules/l4regexp/matcher.go
@@ -1,0 +1,109 @@
+// Copyright 2024 VNXME
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -38,6 +38,7 @@ Current matchers:
 - **layer4.matchers.local_ip** - matches connections based on local IP (or CIDR range).
 - **layer4.matchers.proxy_protocol** - matches connections that start with [HAPROXY proxy protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt).
 - **layer4.matchers.rdp** - matches connections that look like [RDP](https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-RDPBCGR/%5BMS-RDPBCGR%5D.pdf).
+- **layer4.matchers.regexp** - matchers connections that have the first packet bytes matching a regular expression.
```
mholt, line 41:
```suggestion
- **layer4.matchers.regexp** - matches connections that have the first packet bytes matching a regular expression.
```

modules/l4regexp/matcher.go
```diff
@@ -0,0 +1,109 @@
+// Copyright 2024 VNXME
```
mholt, line 1: Typically, throughout the repo, we use the same copyright (a "project" copyright as it were) and leave it to the commit history to assign credit to individual contributors.

That said, I'm not a lawyer nor do I have strong feelings about this in this case in this repo, but I could see if we end up moving this code into Caddy mainline it could be confusing with multiple copyright assignments throughout the code.

I definitely want to credit you though so what if we put a comment below the license statement and say "This file contributed by @vnxme" or something?

mholt (commented):
This looks great! I'm excited to see how people use it.

modules/l4regexp/matcher.go
vnxme: Well, I just copied your typical license block and put my name instead. I think it only indicates that the file is written by me, but it doesn’t restrict any use in the project. Neither was it my intention to put any restrictions.

For example, please see how opnsense copyrights look like [here](https://github.com/opnsense/core/blob/efea5429524e847b0c1371da6c4ff666f54c2459/src/etc/rc#L3). If more than one person contributed to the file, they put multiple names of contributors sorted by years of changes.

mholt: Oh interesting, ok. That's probably fine then.

mholt (approved):
LGTM! Anything else before we merge it?

vnxme:
Rebased onto master and added integration tests. I think, it's ready to be merged now.

mholt:
Awesome! Thanks!

PR #185, opened 2024-04-22
pin quic-go version

IceCodeNew:
Keep up with the upstream.

PR #221, opened 2024-07-24
Disable dependencies updates of gomod

IceCodeNew:
@mohammed90 has mad it clear in this comment https://github.com/mholt/caddy-l4/pull/185#discussion_r1576269100

Instead of updating the dependencies here, we should do it on the upstream ( https://github.com/caddyserver/caddy ), and users are encouraged to keep their caddy up-to-date


mholt (approved):
Great, thanks for the help!

Issue #228, opened 2024-08-05
Cannot proxy to Mariadb instance


KevinBLT:
Hey... maybe I am missing something but since the L4 Module seems to wait for a byte to be sent and is not opening a proxy connection until that moment it will not recieve the mariadb greeting session hence the client also waits until the connection is closed due to a timeout since nobody actually talked.

I tested with telnet:
```bash
kb@MacBook-Air ~ % telnet 127.0.0.1 3306
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
Connection closed by foreign host.

kb@MacBook-Air ~ % telnet 127.0.0.1 3306
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
test
Z
11.2.4-MariaDB-ubu220'0YQjCL7??-j4al|I8{&Fmumysql_native_password5#HY000Proxy header is not accepted from 172.19.0.3Connection closed by foreign host.
```

When I just try to connect... it will close after:
`WARN    layer4  matching connection     {"remote": "192.168.65.1:53364", "error": "aborted matching according to timeout"}`

If I am fast enough to just send random bytes (like here `text`) MariaDB will answer with its greeting but of course terminate because just sent some garbage in its view.

What can I do? Is there a buffer value? Anything related to it?

Caddyfile:
```
{
  layer4 {

    :3306 {
      route {
        proxy {
          upstream mariadb:3306
        }
      }
    }

   
  }
}
```

mholt:
@vnxme Just a quick thought as a side-note: I wonder if instead of `upstream` we should call it `to` like we do in the HTTP proxy (I can't remember if we already talked about this)?

@ydylla Do you think this is fixed by one of the proposed patches? Seems like it shouldn't be waiting to match in this case.

vnxme:
@mholt, I've certainly thought about it. Decided to avoid using `to` and keep `upstream`, because `upstream` here, unlike in the http app, may have multiple inner options like `max_connections` and `tls_*` ones defined *per upstream*. See more [here](https://github.com/mholt/caddy-l4/blob/master/modules/l4proxy/upstream.go#L164-L182).

In this case we could also use a shorter syntax:
```
{
  layer4 {
    :3306 {
      route {
        proxy mariadb:3306
      }
    }
  }
}
```

ydylla:
@mholt 
> Do you think this is fixed by one of the proposed patches?

[...]

PR #229, opened 2024-08-06
fix: make prefetch lazy again

### LINKED ISSUE #212: Regression: PROXY protocol line sent only when first data arrives from client
We have noticed a regression between 145ec36251a4  and ca3e2f38f6e5 when using `proxy_protocol`.

In 145ec36251a4 the `PROXY` protocol line would be sent directly on connection from Caddy to the backend. In ca3e2f38f6e5 it seems the `PROXY` protocol line is sent to the backend only when the client connected to Caddy sends data.

This is problematic with Exim at least, as the SMTP protocol with PROXY (or at least the EXIM implementation of it) expects the mail server/backend to first receive the `PROXY` line, then answers with the version and host information of the mail server/backend, and then expects a `HELO`, `EHLO` or the like. Sending the `PROXY` line with the first line the client sends breaks this dance.

ydylla:
This changes layer4 to again only read lazily from the connection during matching.  
With this tcp proxy configs for protocols where the server speaks first work again.

I have also added some example configs for these cases to #209 "matcher without io" & "server speaks first".

Closes #212, #228 

@mholt This is based on #210 since I wanted to test it together with all the other fixes & tests. I also think this version of the matching loop is even simpler because we do not need a labeled jump or a "global" `routeIdx` anymore.


ydylla:
Superseded by #208

PR #230, opened 2024-08-07
Clock matcher: fix tests in non-UTC timezones

vnxme:
I've found that I missed the UTC part in the tests while making the final adjustments to #225.

With these minor changes, the clock matcher tests won't fail in non-UTC timezones.


IceCodeNew (approved):

mholt (approved):
Ah good catch, thanks!

PR #231, opened 2024-08-09
DNS matcher

vnxme:
### Summary

This PR introduces a DNS protocol matcher. It allows to multiplex plain DNS (TCP and UDP) with other protocols, and DNS-over-TLS (DoT) with HTTPS or basically any TLS-based protocol. Besides, it provides for packet filtering based on the question section of a DNS request message. E.g. you may allow users to query any records of example.com and deny anything else.

This matcher is based on [miekg/dns](https://github.com/miekg/dns) for DNS message parsing and flags validation. This package is also used by [caddyserver/certmagic](https://github.com/caddyserver/certmagic), so the matcher doesn't add new requirements.

Please note that DNS messages sent via TCP are 2 bytes longer then those sent via UDP. Consequently, if Caddy listens on TCP, it has to proxy DNS messages to TCP upstreams only. The same is true for UDP. No TCP/UDP mixing is allowed.

### Syntax

```
dns {
	<allow|deny> <*|name> [<*|type> [<*|class>]]
	<allow_regexp|deny_regexp> <*|name_pattern> [<*|type_pattern> [<*|class_pattern>]]
	default_deny
	prefer_allow
}
```

### Filtering

- An empty `dns` matcher will match any connection if its bytes look like a DNS request message (with the response bit zeroed).
- Multiple `allow`, `allow_regexp`, `deny` and `deny_regexp` options are supported. Each of them introduces a rule to match against the question section of the DNS request message. This section contains a domain name (ending with a dot, e.g. `example.com.`), a record type (e.g. `A` or `MX`) and a class (e.g. `IN`, meaning the Internet). Use an asterisk (`*`) to skip matching any field.
- If only `allow` and `allow_regexp` rules are provided, the implicit *deny all unless allowed* mode is on. In other words, at least one allow rule must match the question section for the matcher to return true.
-  If only `deny` and `deny_regexp` rules are provided, the implicit *allow all unless denied* mode is on. In other words, no deny rule must match the question section for the matcher to return true.
- If both allow and deny rules are provided, there are two helper options to manage the matcher. If the question section of a DNS request message is both allowed and denied by the rules, it is denied by default unless `prefer_allow` is set. If the question section is neither allowed nor denied by the rules, it is allowed by default unless `default_deny` is set.

### Examples

<details>
  <summary>Simple plain DNS proxy</summary>

```
{
	layer4 {
		tcp/:53 {
			@d dns
			route @d {
				proxy tcp/1.1.1.1:53
			}
		}
		udp/:53 {
			@d dns
			route @d {
				proxy udp/1.0.0.1:53
			}
		}
	}
}
```
</details>
<details>
  <summary>DNS-over-TLS (DoT) proxy to a plain DNS upstream</summary>

```
{
	layer4 {
		tcp/:853 {
			@t tls
			route @t {
				tls
				subroute {
					@d dns
					route @d {
						proxy tcp/1.1.1.1:53
					}
				}
			}
		}
	}
}
```
</details>
<details>
  <summary>DNS-over-TLS (DoT) proxy to DoT upstreams</summary>

```
{
	layer4 {
		tcp/:853 {
			@t tls
			route @t {
				tls
				subroute {
					@d dns
					route @d {
						proxy {
							upstream one.one.one.one:853 {
								tls
							}
							upstream 1.0.0.1:853 {
								tls_insecure_skip_verify
							}
						}
					}
				}
			}
		}
	}
}
```
</details>
<details>
  <summary>Filtered plain DNS proxy (deny all unless allowed)</summary>

```
{
	layer4 {
		udp/:53 {
			@d dns {
				# only selected records for example.com and its subdomains are allowed
				allow_regexp ^(|[-0-9a-z]+\\.)example\\.com\\.$ ^(A|AAAA|CNAME|MX|NS|TXT)$ ^IN$
			}
			route @d {
				proxy udp/ns.upstream.local:53
			}
		}
	}
}
```
</details>
<details>
  <summary>Filtered plain DNS proxy (allow all unless denied)</summary>

```
{
	layer4 {
		udp/:53 {
			@d dns {
				# all records for domain.com and its subdomains are denied
				deny_regexp ^(|[-0-9a-z]+\\.)domain\\.com\\.$
			}
			route @d {
				proxy udp/ns.upstream.local:53
			}
		}
	}
}
```
</details>

### Discussion

- As noted above, TCP/UDP mixing (proxy TCP messages to UDP upstream and vice versa) isn't supported. However, it's technically possible: an intermediary handler is required to add/strip 2 bytes before/after proxy. Please write below if you think this feature would be of any use.
- I have a DNS handler WIP based on the same package. My idea is to let Caddy respond to DNS requests as an authoritative server for selected domain names and/or record types rather than be a recursor or a load balancer. But I don't know whether it's worth it. Please share your opinion in the comments.

### DIFF
--- a/modules/l4dns/matcher.go
+++ b/modules/l4dns/matcher.go
@@ -1,0 +1,408 @@
+// Copyright 2024 VNXME
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package l4dns
+
+import (
+	"context"
+	"encoding/binary"
+	"io"
+	"net"
+	"regexp"
+	"strings"
+
+	"github.com/caddyserver/caddy/v2"
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
+	"github.com/miekg/dns"
+
+	"github.com/mholt/caddy-l4/layer4"
+)
+
+func init() {
+	caddy.RegisterModule(&MatchDNS{})
+}
+
+// MatchDNS is able to match connections that look like DNS protocol.
+type MatchDNS struct {
+	// Allow contains an optional list of rules to match the question section of the DNS request message against.
+	// The matcher returns false if not matched by any of them (in the absence of any deny rules).
+	Allow MatchDNSRules `json:"allow,omitempty"`
+	// Deny contains an optional list of rules to match the question section of the DNS request message against.
+	// The matcher returns false if matched by any of them  (in the absence of any allow rules).
+	Deny MatchDNSRules `json:"deny,omitempty"`
+
+	// If DefaultDeny is true, DNS request messages that haven't been matched by any allow and deny rules are denied.
+	// The default action is allow. Use it to make the filter more restrictive when the rules aren't exhaustive.
+	DefaultDeny bool `json:"default_deny,omitempty"`
+	// If PreferAllow is true, DNS request messages that have been matched by both allow and deny rules are allowed.
+	// The default action is deny. Use it to make the filter less restrictive when the rules are mutually exclusive.
+	PreferAllow bool `json:"prefer_allow,omitempty"`
+}
+
+// CaddyModule returns the Caddy module information.
+func (m *MatchDNS) CaddyModule() caddy.ModuleInfo {
+	return caddy.ModuleInfo{
+		ID:  "layer4.matchers.dns",
+		New: func() caddy.Module { return new(MatchDNS) },
+	}
+}
+
+// Match returns true if the connection bytes represent a valid DNS request message.
+func (m *MatchDNS) Match(cx *layer4.Connection) (bool, error) {
+	var (
+		msgBuf   []byte
+		msgBytes uint16
+	)
+
+	// Detect the connection protocol: TCP or UDP.
+	// Note: all non-TCP connections are treated as UDP, so no TCP packets could be matched while testing
+	// with net.Pipe() unless a valid cx.LocalAddr() response is provided using a fakeTCPConn wrapper.
+	if _, ok := cx.LocalAddr().(*net.TCPAddr); ok {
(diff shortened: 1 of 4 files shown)


modules/l4dns/matcher.go
```diff
@@ -0,0 +1,408 @@
...
+	"github.com/miekg/dns"
+
+	"github.com/mholt/caddy-l4/layer4"
+)
+
+func init() {
+	caddy.RegisterModule(&MatchDNS{})
+}
+
+// MatchDNS is able to match connections that look like DNS protocol.
```
mholt, line 36: Should probably document somewhere that if proxying DNS connections, TCP/UDP can't be mixed from inbound to outbound (but to file an issue if this is needed).

mholt (approved):
Very cool!!

> As noted above, TCP/UDP mixing (proxy TCP messages to UDP upstream and vice versa) isn't supported. However, it's technically possible: an intermediary handler is required to add/strip 2 bytes before/after proxy. Please write below if you think this feature would be of any use.

Let's wait for a legitimate feature request for this. Definitely a good caveat to note in the docs in the meantime though.

> I have a DNS handler WIP based on the same package. My idea is to let Caddy respond to DNS requests as an authoritative server for selected domain names and/or record types rather than be a recursor or a load balancer. But I don't know whether it's worth it. Please share your opinion in the comments.

Essentially some of the basic functionality of CoreDNS? I've always thought CoreDNS would make a good Caddy v2 app module. (It started as a fork of Caddy v1. Never did get updated to Caddy v2.) Anyway, I think a DNS app would make a great module for Caddy. Maybe a DNS handler for this app could be a starting point, and then later a DNS app could reuse that same code/logic.

Anyway, this feature is really cool. I think it will be useful! Anything left before I merge this?

modules/l4dns/matcher.go
vnxme: OK, done.

vnxme:
> Essentially some of the basic functionality of CoreDNS? I've always thought CoreDNS would make a good Caddy v2 app module. (It started as a fork of Caddy v1. Never did get updated to Caddy v2.) Anyway, I think a DNS app would make a great module for Caddy. Maybe a DNS handler for this app could be a starting point, and then later a DNS app could reuse that same code/logic.

Yes, but I've started with the same [miekg/dns](https://github.com/miekg/dns) package. Will look into CoreDNS as well.

> Anything left before I merge this?

No, unless there are any objections/mistakes/etc.

mholt:
Sounds good.

I would stick to the miekg/dns package, mainly because CoreDNS I think is more built as a standalone application, rather than a library. If the miekg/dns package does what you need, then that would be ideal I think. I just referred to CoreDNS in terms of its set of functionality that I think could transfer very well to a Caddy v2 app.

Thanks, this is cool!

Issue #234, opened 2024-08-15
Incorrect Handling of SNI: No Certificate Available (Regression)


elcore:
After updating the Caddy Layer4 module to https://github.com/mholt/caddy-l4/commit/e491c44895fe3f11e24ad4d8c4f6a668144c0ef9, Caddy erroneously attempts to handle certificates for domains when it should instead proxy and forward connections based on SNI.

Logs:

```
{"level":"error","ts":1723754927.6938345,"logger":"layer4","msg":"handling connection","remote":"127.0.0.1:35770","error":"no certificate available for 'ud-chat.signal.org'"}
{"level":"error","ts":1723754927.7430856,"logger":"layer4","msg":"handling connection","remote":"127.0.0.1:35772","error":"no certificate available for 'chat.signal.org'"}
{"level":"error","ts":1723754927.9744139,"logger":"layer4","msg":"handling connection","remote":"127.0.0.1:35788","error":"no certificate available for 'ud-chat.signal.org'"}
{"level":"error","ts":1723754928.1224709,"logger":"layer4","msg":"handling connection","remote":"127.0.0.1:35802","error":"no certificate available for 'chat.signal.org'"}
{"level":"error","ts":1723754929.6092706,"logger":"layer4","msg":"handling connection","remote":"127.0.0.1:60324","error":"no certificate available for 'ud-chat.signal.org'"}
...
```

Configuration: 
```
{
  "apps": {
    "tls": {
      "certificates": {
        "automate": [
          "sub.example.com"
        ]
      }
    },
    "layer4": {
      "servers": {
        "signal_proxy": {
          "listen": [
            ":443"
          ],
          "routes": [
            {
              "handle": [
                {
                  "handler": "tls"
                }
              ]
            },
            {
              "match": [
                {
                  "tls": {
                    "sni": [
                      "chat.signal.org",
                      "ud-chat.signal.org",
                      "storage.signal.org",
                      "cdn.signal.org",
                      "cdn2.signal.org",
                      "cdn3.signal.org",
                      "api.directory.signal.org",
                      "cdsi.signal.org",
                      "sfu.voip.signal.org",
                      "svr2.signal.org",
                      "updates.signal.org",
                      "updates2.signal.org",
                      "backend1.svr3.signal.org",
                      "backend2.svr3.signal.org",
                      "backend3.svr3.signal.org"
                    ]
                  }
                }
              ],
              "handle": [
                {
                  "handler": "proxy",
                  "upstreams": [
[...]

PR #235, opened 2024-08-16
Fix regression in #224: allow dynamic upstreams

vnxme:
This is a fix addressing the issue described in https://github.com/mholt/caddy-l4/issues/234.


elcore:
Is it alright for you if I test this on Monday? I don’t have time over the weekend

elcore:
It works

mohammed90 (approved):
Great. Thanks, @elcore for the report and confirmation. Thanks @vnxme for the fix!

PR #236, opened 2024-08-16
Adjust matchers after #208: return reading errors

vnxme:
This is a fix adjusting selected matchers to return reading errors so that the matching loop could continue after #208.


mholt (approved):
Thanks! Sorry for my delay :)

Issue #247, opened 2024-09-28
UDP gets blocked after any unsuccessful matching


vnxme:
I'm currently debugging an OpenVPN matcher, and I've noticed a problem with how UDP is handled by caddy-l4.

### Symptoms:
- A UDP connection gets blocked / hangs, i.e. no packets are passed for matching, after the first unsuccessful matching.
- If an invalid packet is sent to Caddy after a valid packet, then any valid packets that follow are completely ignored.
- The problem always persists until Caddy is restarted.

### Sample config:
```
{
	layer4 {
		udp/:53 {
			@test dns {
				allow apple.com. * *
				default_deny
			}
			route @test {
				proxy udp/1.1.1.1:53
			}
		}
	}
}
```

### Replication:
- The problem is related to any UDP packets, but I've found DNS to be the easiest way to replicate. Send any non-apple.com DNS request to Caddy running the sample config above, then send apple.com DNS request - it won't respond. Restart Caddy and send apple.com DNS request - it will respond, then send any non-apple.com DNS request, then send apple.com DNS request again - you will get no response.
- If `fmt.Printf("%x,  %d\n", buf[:10], n)` is inserted before https://github.com/mholt/caddy-l4/blob/master/layer4/server.go#L109, it never prints anything after unsuccessful matching.

### Side problem:
- When I do UDP debugging, Caddy panics on https://github.com/mholt/caddy-l4/blob/master/layer4/server.go#L156 from time to time. I think it depends on how much time has passed between the app stops on a breakpoint and I press the resume button.

vnxme:
I think the issue described above can be traced to https://github.com/mholt/caddy-l4/commit/6a8be7c4b8acb0c531b6151c94a9cd80894acce1.

@jtackaberry Is there any chance you could look at it?

vnxme:
Actually it turns out to be the matchingTimeout set by https://github.com/mholt/caddy-l4/blob/master/layer4/routes.go#L123 that breaks UDP.

As a workaround (and a proof that the timeout is at least related to the problem described above), we can put `_ = conn.SetReadDeadline(time.Time{})` as a simple neutralisation command after https://github.com/mholt/caddy-l4/blob/master/layer4/server.go#L147, and it seems to effectively prevent UDP from dying after unsuccessful matching. 

[...]

PR #249, opened 2024-10-02
set individual deadline when matching udp connection

WeidiDeng:
fix [247](https://github.com/mholt/caddy-l4/issues/247).


WeidiDeng:
An actual read deadline is implemented using timers.

vnxme:
Wow, thank you for the additional effort you've contributed to this issue!

I will re-test the OpenVPN matcher and confirm if your extended fix works with it as fine as the original version.

mholt (approved):
Thanks so much for the patch, and for the verification @vnxme !

PR #251, opened 2024-10-02
OpenVPN matcher

vnxme:
### Summary

This PR introduces an advanced OpenVPN matcher capable of:
- matching any of the existing control channel security modes (plain, auth, crypt and crypt2);
- matching based on a digest algorithm (auth mode);
- matching based on a group key (auth and crypt modes);
- matching based on a server key (crypt2 mode);
- matching based on client keys (crypt2 mode);
- matching both TCP and UDP connections.

These features allow Caddy to multiplex a plethora of diversely configured OpenVPN server instances on a single port. It is a unique OpenVPN routing solution that goes beyond what is offered by SSLH, Haproxy, Nginx, Traefik and OpenVPN internal port-sharing.

### Sample config

```
{
	layer4 {
		:8843 {
			@plain openvpn {
				modes plain
			}
			route @plain {
				proxy localhost:1194
			}
			@auth_sha256 openvpn {
				modes auth
				auth_digest sha256
				group_key_direction normal
				group_key_file /etc/openvpn/ta.key
			}
			route @auth_sha256 {
				proxy localhost:1195
			}
			@auth_sha512 openvpn {
				modes auth
				auth_digest sha512
				group_key_direction inverse
				group_key_file /etc/openvpn/ta.key
			}
			route @auth_sha512 {
				proxy another.machine.local:1295
			}
			@crypt_key1 openvpn {
				modes crypt
				group_key_file /etc/openvpn/ta.key
			}
			route @crypt_key1 {
				proxy localhost:1196
			}
			@crypt_key2 openvpn {
				modes crypt
				group_key_file /etc/openvpn/ta2.key
			}
			route @crypt_key2 {
				proxy another.machine.local:1296
			}
			@crypt2_server openvpn {
				modes crypt2
				server_key_file /etc/openvpn/v2-server.key
			}
			route @crypt2_server {
				proxy localhost:1197
			}
			@crypt2_clients openvpn {
				modes crypt2
				client_key_file /etc/openvpn/v2-client1.key
				client_key_file /etc/openvpn/v2-client2.key
				client_key_file /etc/openvpn/v2-client3.key
			}
			route @crypt2_clients {
				proxy another.machine.local:1297
			}
			route {
				tls
				proxy localhost:8080
			}
		}
		udp/:8843 {
			@anymode openvpn {
				modes plain auth crypt crypt2
			}
			route @anymode {
				proxy udp/localhost:1199
			}
			route {
				echo
			}
		}
	}
}
```

### Disclaimer

- I have tested this matcher with OpenVPN Server 2.6.8 (Linux), OpenVPN Client 3.5.0 (Windows), Viscosity 1.10.2 (macOS). Please share details and your config should you encounter any issues.
- As documented in the code, digest support (auth mode) is limited to a number of Golang's standard ones, i.e. some exotic algorithms like SM3 and Whirlpool are not (yet?) included. There is a special case of MD5+SHA1 which is implemented by this matcher but its HMACs don't match those of OpenVPN clients I tested.
- OpenVPN has to renegotiate connections on Caddy restart. For UDP, after the last packet the client received from the server, up to 120 seconds (default renegotiation interval) may pass before a previously established connection goes live again. TCP connections are usually renegotiated much faster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ Current matchers:
 - **layer4.matchers.http** - matches connections that start with HTTP requests. In addition, any [`http.matchers` modules](https://caddyserver.com/docs/modules/) can be used for matching on HTTP-specific properties of requests, such as header or path. Note that only the first request of each connection can be used for matching.
 - **layer4.matchers.local_ip** - matches connections based on local IP (or CIDR range).
 - **layer4.matchers.not** - matches connections that aren't matched by inner matcher sets.
+- **layer4.matchers.openvpn** - matches connections that look like [OpenVPN](https://openvpn.net/community-resources/openvpn-protocol/) connections.
 - **layer4.matchers.postgres** - matches connections that look like Postgres connections.
 - **layer4.matchers.proxy_protocol** - matches connections that start with [HAPROXY proxy protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt).
 - **layer4.matchers.rdp** - matches connections that look like [RDP](https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-RDPBCGR/%5BMS-RDPBCGR%5D.pdf).
--- a/imports.go
+++ b/imports.go
@@ -21,6 +21,7 @@ import (
 	_ "github.com/mholt/caddy-l4/modules/l4dns"
 	_ "github.com/mholt/caddy-l4/modules/l4echo"
 	_ "github.com/mholt/caddy-l4/modules/l4http"
+	_ "github.com/mholt/caddy-l4/modules/l4openvpn"
 	_ "github.com/mholt/caddy-l4/modules/l4postgres"
 	_ "github.com/mholt/caddy-l4/modules/l4proxy"
 	_ "github.com/mholt/caddy-l4/modules/l4proxyprotocol"
--- a/integration/caddyfile_adapt/gd_matcher_openvpn.caddytest
+++ b/integration/caddyfile_adapt/gd_matcher_openvpn.caddytest
@@ -1,0 +1,169 @@
+{
+	layer4 {
+		:8843 {
+			@plain openvpn {
+				modes plain
+			}
+			route @plain {
+				proxy localhost:1194
+			}
+			@auth openvpn {
+				modes auth
+				auth_digest sha256
+				group_key_direction normal
+				group_key_file /etc/openvpn/ta.key
+			}
+			route @auth {
+				proxy localhost:1195
+			}
+			@crypt openvpn {
+				modes crypt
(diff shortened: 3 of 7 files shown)


mholt:
_Wow, this is impressive!_

Looks like this may require Go 1.22. That's OK if so, I just want to verify that is intentional.

I have not the ability to test this myself, but I am happy to merge this in for others to test, when you are ready. :100: Thank you!

vnxme:
> Looks like this may require Go 1.22. That's OK if so, I just want to verify that is intentional.

It is slices.Concat that prevents compiling with Go 1.21. It is used in one place though, so I can rewrite it. I wouldn't call it intentional, but forward-looking - I don't know a reason for sticking to old versions here given the mainline code already requires Go 1.22.

By the way, what do you think of switching to Go 1.22/1.23 in caddy-l4 go.mod? I suppose we would also have to update checks then.

mholt:
Ah, right, I forgot that 1.21 is 2 versions old now. Yeah, we can upgrade the Go versions in the tests and Go.mod to accommodate the slices package. :+1: Want to do that in this PR?

vnxme:
I've prepared https://github.com/mholt/caddy-l4/pull/253. I will rebase this PR once you approve and merge that dependencies upgrade.

I suggest we merge this PR only after https://github.com/mholt/caddy-l4/issues/247 is resolved (whether with https://github.com/mholt/caddy-l4/pull/249 that works fine for me or anything else). Otherwise the first users of this matcher won't be happy with their broken UDP-based OpenVPN tunnels.

mholt (approved):
Amazing. Thanks so much -- I've merged the PR with the patch, so I will merge this now! :rocket:

Juoper:
@vnxme trying to figure out the different modes you made available, when i want all my openvpn traffic to be routed to 10.10.10.10 could i just use this config:

{
    layer4 {
        :443 {
                          @plain openvpn {
				modes plain
			 }
			route @plain {
				proxy 10.10.10.10:443
			}
            }
        }
    }
}

vnxme:
@Juoper Does it work if you don’t set `modes` at all (comment out this line)? If you use UDP, modify your listener accordingly. Besides, make sure you don’t serve any https websites on the same port - if you do, then you will need a listener_wrapper for TCP traffic (and disable QUIC if you also need UDP).

Juoper:
i serve https on the same port, could you assist me to tell me how i can implement a listener_wrapper

vnxme:
@Juoper The following config will likely satisfy your needs.
```
{
	servers {
		listener_wrappers {
			layer4 {
				@ovpn openvpn
				route @ovpn {
					# note that a listener wrapper will proxy only TCP connections
					# since the current listener_wrappers implementation ignores UDP
					proxy tcp/10.10.10.10:443
				}
			}
			tls
		}
	}
}

example.com {
	# your options
}
```

Juoper:
Thanks a lot, this is working now, maybe someone could add some example like this to the docs?

vnxme:
@Juoper The caddy-l4 module is very flexible, and it's barely possible to describe all the combinations of the matchers and handlers it includes. But feel free to submit a PR and/or add examples to the wiki.

PR #255, opened 2024-10-12
Winbox matcher

vnxme:
### Summary

This PR introduces a new matcher to multiplex connections initiated by [Winbox](https://help.mikrotik.com/docs/display/ROS/WinBox), a graphical tool for MikroTik hardware and software routers management. Many of these routers all over the Internet have port 8291 open and listening for incoming connections on their WANs. Some of them allegedly formed part of a global bot net a few years ago, and a few Winbox vulnerabilities were most probably the reason for it.

This matcher allows to multiplex Winbox connections with anything else (e.g. TLS on port 443), so that such routers could be reachable without exposing their distinctive port. The matcher is also capable of filtering users (with plaintext or regular expressions) and modes (standard and/or RoMON), so that no unwanted connection requests could reach the router(s) behind Caddy.

Given RouterOS is a commercial software and Winbox uses an undocumented proprietary protocol to connect to it, proxying Winbox through Caddy could be an additional layer of security to mitigate vulnerabilities that may already exist or be introduced in the future.

### Sample syntax
```
{
	layer4 {
		:443 {
			@w1 winbox {
				modes standard romon
				username toms
			}
			route @w1 {
				proxy 192.168.0.1:8291
			}
			@w2 winbox {
				modes standard
				username_regexp ^andris|edgars|juris$
			}
			route @w2 {
				proxy 192.168.0.2:8291
			}
			@w3 winbox
			route @w3 {
				proxy 192.168.0.3:8291
			}
		}
	}
}
```

### Disclaimer

I've tested this matcher against Winbox v3.30, v3.41 and v4.0 in both standard and RoMON modes. A legacy mode which is available in the tools menu of Winbox didn't make any difference in terms of matching the first packet received from the client, i.e. it is also inherently supported. The previous versions of Winbox used to have a secure mode checkbox, and this matcher will most probably only work when it is checked. The newer versions have it implicitly enabled by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Current matchers:
 - **layer4.matchers.socks5** - matches connections that look like [SOCKSv5](https://www.rfc-editor.org/rfc/rfc1928.html).
 - **layer4.matchers.ssh** - matches connections that look like SSH connections.
 - **layer4.matchers.tls** - matches connections that start with TLS handshakes. In addition, any [`tls.handshake_match` modules](https://caddyserver.com/docs/modules/) can be used for matching on TLS-specific properties of the ClientHello, such as ServerName (SNI).
+- **layer4.matchers.winbox** - matches connections that look like those initiated by [Winbox](https://help.mikrotik.com/docs/display/ROS/WinBox), a graphical tool for MikroTik hardware and software routers management.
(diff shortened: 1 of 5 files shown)


mholt (approved):
Very neat, thanks for contributing this!

I don't know anything about this protocol, but it looks like you do :smile:

Ghibranalj:
WOW this is amazing!!! im soo glad i found this. GREAT WORK man 👏🏻

Issue #259, opened 2024-10-25
Panic with match tls alpn xmpp-client 


tctlrd:
I have started to get a panic in caddy w/L4 from part of a config for an XMPP server that matches for alpn: xmpp-client.
It was working previously, but I updated caddy and now it seems to have broken or changed. I also tried with a json config with no yaml adapter and had the same issue.
The panic happens as soon as I visit a page that caddy is serving.
Is anyone else experiencing this?

I removed these lines and caddy is fine:
https://github.com/snikket-im/snikket-server/blob/f89e01402b9f5d5f63c9c54152531c3755587abd/docs/advanced/reverse_proxy.md?plain=1#L380-L389
```
routes: 
 - match: 
   - tls:      # match encrypted XMPP traffic 
       alpn: 
       - xmpp-client 
   handle: 
   - handler: proxy 
     upstreams: 
     - dial:   # and send to Snikket's encrypted XMPP port 
       - localhost:5223 
```

Here is the panic:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x159e52e]

goroutine 43 [running]:
github.com/mholt/caddy-l4/modules/l4tls.(*MatchALPN).Match(0xc000699d70, 0xc0007a4488)
	github.com/mholt/caddy-l4@v0.0.0-20240912211411-4f012d4517cf/modules/l4tls/alpn_matcher.go:40 +0x2e
github.com/mholt/caddy-l4/modules/l4tls.(*MatchTLS).Match(0xc000745380, 0xc0003387e0)
	github.com/mholt/caddy-l4@v0.0.0-20240912211411-4f012d4517cf/modules/l4tls/matcher.go:115 +0x2ea
github.com/mholt/caddy-l4/layer4.MatcherSet.Match({0xc0007398a0?, 0x596325?, 0xc00069ace8?}, 0xc0003387e0)
	github.com/mholt/caddy-l4@v0.0.0-20240912211411-4f012d4517cf/layer4/matchers.go:53 +0xa3
github.com/mholt/caddy-l4/layer4.(*MatcherSets).AnyMatch(0xc0006336b0, 0xc0003387e0)
	github.com/mholt/caddy-l4@v0.0.0-20240912211411-4f012d4517cf/layer4/matchers.go:88 +0x6f
[...]

PR #260, opened 2024-10-25
Fix a regression in #224: the context may have no replacer

vnxme:
This PR fixes the panics caused by the `tls alpn` matcher as reported in #259. The fix is a copy of https://github.com/caddyserver/caddy/commit/7cf8376e638948490be3e9eb5c7d58ce2a4b93b3 we merged into the mainline for the `tls sni` matcher a few month ago.

### DIFF
--- a/modules/l4tls/alpn_matcher.go
+++ b/modules/l4tls/alpn_matcher.go
@@ -37,7 +37,14 @@ func (*MatchALPN) CaddyModule() caddy.ModuleInfo {
 }
 
 func (m *MatchALPN) Match(hello *tls.ClientHelloInfo) bool {
-	repl := hello.Context().Value(caddy.ReplacerCtxKey).(*caddy.Replacer)
+	repl := caddy.NewReplacer()
+	if ctx := hello.Context(); ctx != nil {
+		// In some situations the existing context may have no replacer
+		if replAny := ctx.Value(caddy.ReplacerCtxKey); replAny != nil {
+			repl = replAny.(*caddy.Replacer)
+		}
+	}
+
 	clientProtocols := hello.SupportedProtos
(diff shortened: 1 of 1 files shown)


mholt:
Thanks -- kind of unfortunate we can't reliably know there's a Replacer available. Do we know if it's rather possible to ensure a Replacer is always in the context? I wonder if that would be easier than ensuring every handler/matcher does this check.

vnxme:
Well, my understanding is as follows:
- As fairly mentioned [here](https://github.com/mholt/caddy-l4/blob/16b3b200e1828ec0ad7cfdf88bf67b3a0c80cda0/modules/l4tls/matcher.go#L109-L114), the tls app matchers only accept tls.ClientHelloInfo. We do have a valid replacer in a layer4.Connection, but we can't pass it through.
- Given we fill the tls.ClientHelloInfo [here](https://github.com/mholt/caddy-l4/blob/16b3b200e1828ec0ad7cfdf88bf67b3a0c80cda0/modules/l4tls/matcher.go#L100), I expect its ctx field to always equal nil, since we don't assign any value to it because it isn't exported.
- I assume we don't use this alpn matcher in the mainline code, but if we did, its ctx field would most probably have a valid replacer since we pass the app context to the tls.Config (unless we fake a real connection in tests, e.g. caddytls.TestServerNameMatcher).

So I would conclude that:
- For layer4 purposes tls.ClientHelloInfo doesn't have a context, that's why we create a new replacer;
- For mainline purposes (if we use this matcher anywhere else) tls.ClientHelloInfo will have a context, and we try to reuse the existing matcher, if any.

mholt (approved):
Good points. Thanks for this patch.

Issue #258, opened 2024-10-25
Caddy will not gracefully stop when SSH is proxied


Monviech:
**Caddy will hang indefinitely on stop with this setup:**

https://docs.opnsense.org/manual/how-tos/caddy.html#ssh-multiplexing-on-https-port

**User report:**

https://forum.opnsense.org/index.php?topic=38714.msg217113#msg217113

**My own minimal test case to reproduce it (on FreeBSD 14.1):**

caddy-l4 version used: https://github.com/mholt/caddy-l4/commit/bdee6a6620c02aca4817e6c3091fed26aa10940c

```
{
	http_port 8080
	https_port 8443

	servers {
		listener_wrappers {
			layer4 {

				@f9088005-43c9-404d-af09-41fae900ed03 ssh
				route @f9088005-43c9-404d-af09-41fae900ed03 {
					proxy tcp/127.0.0.1:22 {
					}
				}
				route
			}
			tls
		}
	}
	grace_period 10s
}

:8080 {
}

:8443 {
}
```

Open SSH connection: `ssh root@172.16.1.109 -p 8443`

Try to stop caddy: `service caddy stop`

Caddy will hang indefinitely or until the SSH connection is disconnected. The `grace_period` will not kill this context for layer4 proxy.

**Expected behavior:**

Any Layer 4 proxy will adhere to the `grace_period` on SIGTERM and disconnect all connections so caddy can stop.

mholt:
One thing to note, unless it's in the `layer4` config block, the rest of the config applies to HTTP servers.

But I'll try to find out why it's not closing connections.

Monviech:
Its using the admin interface to tell caddy to stop.

https://github.com/opnsense/ports/blob/0a04fb51826bb9102ee3c9b9704138837ae4444d/www/caddy-custom/files/caddy.in#L148

Monviech:
Here's the logfiles. It will hang indefinitely after
```
2024-10-25T19:20:06 | Informational | caddy | "info","ts":"2024-10-25T19:20:06Z","logger":"http","msg":"servers shutting down; grace period initiated","duration":10}
```
Full log from initial reload till stop while SSH is active:

```
2024-10-25T19:20:06	Informational	caddy	 "info","ts":"2024-10-25T19:20:06Z","logger":"http","msg":"servers shutting down; grace period initiated","duration":10}
[...]

PR #261, opened 2024-10-28
stop the loop when listener is closed

WeidiDeng:
Fix [258](https://github.com/mholt/caddy-l4/issues/258).

### DIFF
--- a/layer4/listener.go
+++ b/layer4/listener.go
@@ -134,13 +134,8 @@ func (l *listener) Close() error {
 func (l *listener) loop() {
 	for {
 		conn, err := l.Listener.Accept()
-		// listener closed
-		if l.closed.Load() {
-			break
-		}
-
 		var nerr net.Error
-		if errors.As(err, &nerr) && nerr.Temporary() {
+		if errors.As(err, &nerr) && nerr.Temporary() && !l.closed.Load() {
 			l.logger.Error("temporary error accepting connection", zap.Error(err))
 			continue
 		}
--- a/layer4/listener.go
+++ b/layer4/listener.go
@@ -125,9 +125,8 @@ type listener struct {
 }
 
 func (l *listener) Close() error {
-	err := l.Listener.Close()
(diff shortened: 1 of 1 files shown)


mholt:
Thanks Weidi! As usual you beat me to it. Do you know why this is needed though? Like why doesn't closing the listener bump out of the Accept loop? (Only when SSH connections are active...?)

WeidiDeng:
The listener wrapper is used in my own listener plugin, which is only used on linux. I guess because listeners on linux use port reuse by default, the error there was temporary. On other platforms, it's implemented via a deadline, which is temporary.

WeidiDeng:
It's not ssh though, the listener loop is still running, and the error [here](https://github.com/golang/go/blob/master/src/net/http/server.go#L3334) is never received, because this listener silently discard it.

layer4/listener.go
```diff
@@ -114,25 +116,35 @@ type listener struct {
 	logger        *zap.Logger
 	compiledRoute Handler
 
+	closed atomic.Bool
```
mholt, line 119: Would a [sync.Once](https://pkg.go.dev/sync#Once) be simpler here?

mholt (commented):
Thanks for working on this!

Just one idea to make things a bit simpler...

layer4/listener.go
WeidiDeng: `sync.Once` is used to make sure something happens once, not used to test if something happened. It doesn't fit in the struct

mholt: Ah, I see below, the done channel is used for signaling.

Monviech:
Should I test this again or is it ready?

mholt:
I'll circle back to this soon! Been busy

Monviech:
@mholt No stress just wanted to ask <3

mholt (approved):

mholt:
CI failure seems unrelated

PR #262, opened 2024-10-31
QUIC matcher

vnxme:
### Summary

This PR introduces a `quic` matcher that enables Caddy to multiplex QUIC with anything else like OpenVPN and Wireguard on a single UDP port. This matcher supports all the currently existing `tls` matchers, including `alpn`, `sni`, `local_ip`, and `remote_ip`. Besides, it exposes the following placeholders: `{l4.quic.tls.server_name}`, `{l4.quic.tls.version}`, and `{l4.quic.version}`.

### Syntax
```
{
	layer4 {
		udp/:8443 {
			@q1 quic sni one.com two.com
        		route @q1 {
        			proxy udp/{l4.quic.tls.server_name}:443
        		}
			@q2 quic {
				alpn custom
				sni example.com
			}
        		route @q2 {
        			proxy udp/localhost:6543
        		}
		}
	}
}
```

### Discussion / Further steps
- I can see 3 major places where it would be great to implement QUIC support consistent with what we have now for TLS: 
  1/ a `quic` handler, i.e. an intermediary unwrapper to handle incoming QUIC connections;
  2/ QUIC support in the `proxy` handler, i.e. an intermediary wrapper to handle outgoing QUIC connections;
  3/ QUIC, and generally UDP, wrapper in the mainline code, i.e. a listener_wrappers alternative for net.PacketConn connections.
- As far as the first 2 steps are concerned, I've found an [example of QUIC proxy](https://github.com/atoonk/tcp-to-quic-proxy/blob/main/internal/proxy/proxy.go). However, as of now I don't fully understand how to make it work correctly within the existing caddy-l4 architecture. Any help would be appreciated.


mholt (approved):
Legend.

This is great :smile: 

To implement a QUIC proxy like that, it looks like the listener itself needs to be modified... but QUIC is just UDP. I'd be curious if we can "drop a layer" lower in the QUIC library to be able to use a plain UDP listener, and sprinkle QUIC on top.

vnxme:
I’ve identified an issue of the matcher not working correctly with [curl](https://github.com/curl/curl) (though it works fine with [quicreach](https://github.com/microsoft/quicreach) in my tests). I’m investigating it and will likely have to fix something.

PR #263, opened 2024-11-02
QUIC matcher: curl compatibility and misc changes

vnxme:
Following #262, I've identified that [curl](https://github.com/curl/curl) (unlike [quicreach](https://github.com/microsoft/quicreach)):
- doesn't like/support ED25519 as a private key algorithm for a QUIC server certificate, but works fine with RSA and ECDSA;
- doesn't receive correct responses from [quic-go](https://github.com/quic-go/quic-go) when the latter has a `tls.Certificate` with the `Leaf` field set and the `Certificate` field unset (when it is properly filled, the `Leaf` field is parsed automatically).

I've made a few more minor updates to the code and added extra tests to address curl compatibility.


mholt (approved):
Cool. LGTM. Thanks a bunch!

PR #266, opened 2024-11-05
L4 module "remote_ip_list" to support fail2ban

trefzaxSICKAG:
This PR provides a module to integrate fail2ban into L4 and works similar to https://github.com/Javex/caddy-fail2ban.
The module can be used to protect protocols on layer 4 against brute force attacks by allowing to block an IP address after too many failed login attempts. I have used it to protect a MQTT server, but it could also be used for other protocols with login mechanisms like FTP, SSH, OPC UA, SNMP, IMAP, POP3, SMTP, LDAP...

The module introduces a matcher `remote_ip_list` which can be used to check if the IP address of the sender is contained in a remote IP list. This can be used to e.g. ban the matched IP.

Example Caddyfile:
```
layer4 {
    :1883 {
        @notbanned {
            not {
                remote_ip_list banned-ips
            }
        }
        route @notbanned {
            proxy {
                proxy_protocol v1
                upstream message-broker:1883
            }
        }
    }
}
```


modules/l4fail2ban/matcher.go
```diff
@@ -0,0 +1,130 @@
...
+	// Create new banlist, same as in http.matchers.fail2ban (https://github.com/Javex/caddy-fail2ban/blob/main/banlist.go)
+	m.banlist = caddy_fail2ban.NewBanlist(ctx, m.logger, &m.Banfile)
+	m.banlist.Start()
+	return nil
+}
+
+// The Match will return true if the remote IP is found in the ban list
+func (m *Fail2Ban) Match(cx *layer4.Connection) (bool, error) {
+	clientIP, err := m.getRemoteIP(cx)
+	strClientIP := clientIP.String()
```
vnxme, line 62: I suppose there is no point in converting clientIP into string if err != nil. So I would move this line down, just before the IsBanned() condition.

trefzaxSICKAG: Good catch, I just moved the line down.

modules/l4fail2ban/matcher.go
```diff
@@ -0,0 +1,129 @@
...
+	return ip, nil
+}
+
+// UnmarshalCaddyfile sets up the banfile_path from Caddyfile. Syntax:
+//
+// fail2ban <banfile_path>
+func (m *Fail2Ban) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	_, wrapper := d.Next(), d.Val() // consume wrapper name
+
+	if d.Val() != "fail2ban" {
```
vnxme, line 103: I would replace d.Val() with wrapper in this line.

modules/l4fail2ban/matcher_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+// Cleanup the temporary directory and the ban file
+func cleanupBanFile(t *testing.T, tempDir string) {
+	t.Helper()
+	err := os.RemoveAll(tempDir)
+	assertNoError(t, err)
+}
+
+// Test the Caddyfile unmarshaller
+func TestFail2BanUnmarshaller(t *testing.T) {
```
vnxme, line 87: Why don't you just add an integration test? See an [example](https://github.com/mholt/caddy-l4/blob/master/integration/caddyfile_adapt/gd_matcher_regexp.caddytest).

go.mod
```diff
@@ -5,9 +5,11 @@ go 1.22.0
 toolchain go1.23.0
 
 require (
+	github.com/Javex/caddy-fail2ban v0.0.0-20240525135356-cbf9f5a003ce
 	github.com/caddyserver/caddy/v2 v2.8.4
 	github.com/mastercactapus/proxyprotocol v0.0.4
 	github.com/miekg/dns v1.1.62
+	github.com/quic-go/quic-go v0.44.0
```
vnxme, line 12: Seems like I failed to update go.mod within one of the latest QUIC PRs. That's why it's visible as an addition here.

vnxme:
Hi! I actually like the idea of this matcher. As far as I understand, it requires https://github.com/Javex/caddy-fail2ban (a realtively new plugin I haven't conducted a deep analysis of, but it will be inside a go binary, so it's not a big deal then) and https://github.com/fail2ban/fail2ban (a completely separate non-go project). In other words, even if Caddy is built with caddy-l4 and this matcher included, the desired fail2ban functionality won't work unless one installs and configures fail2ban separately. Thus, it makes me doubt whether this matcher should be part of caddy-l4, https://github.com/Javex/caddy-fail2ban, or even be a separate solution.

modules/l4fail2ban/matcher.go
vnxme: I guess UnmarshalCaddyfile() is only called by Caddy when it matches the beginning of the config line to the last word of what you have in CaddyModule(). Why do you think this condition is required?

modules/l4fail2ban/matcher.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/mholt/caddy-l4/layer4"
+	"go.uber.org/zap"
+)
+
+func init() {
+	caddy.RegisterModule(&Fail2Ban{})
+}
+
+type Fail2Ban struct {
+	Banfile string `json:"banfile"`
```
mohammed90, line 35: It's more idiomatic in Go to have Pascal case for struct field and snake case for json fields

```suggestion
	BanFile string `json:"ban_file"`
```

modules/l4fail2ban/matcher.go
```diff
@@ -0,0 +1,129 @@
...
+		New: func() caddy.Module { return new(Fail2Ban) },
+	}
+}
+
+// Provision implements caddy.Provisioner.
+func (m *Fail2Ban) Provision(ctx caddy.Context) error {
+	m.logger = ctx.Logger()
+
+	// Create new banlist, same as in http.matchers.fail2ban (https://github.com/Javex/caddy-fail2ban/blob/main/banlist.go)
+	m.banlist = caddy_fail2ban.NewBanlist(ctx, m.logger, &m.Banfile)
```
mohammed90, line 54:
```suggestion
	m.banlist = caddy_fail2ban.NewBanlist(ctx, m.logger, &m.BanFile)
```

modules/l4fail2ban/matcher.go
```diff
@@ -0,0 +1,129 @@
...
+		return d.ArgErr()
+	}
+
+	// Only one same-line argument is supported
+	if d.CountRemainingArgs() > 1 {
+		return d.ArgErr()
+	}
+
+	if d.NextArg() {
+		m.Banfile = d.Val()
```
mohammed90, line 113:
```suggestion
		m.BanFile = d.Val()
```

mohammed90 (changes requested):
One thing stood out to me. I couldn't capture all the references as I'm trying to do it on the phone.

go.mod
```diff
@@ -5,9 +5,11 @@ go 1.22.0
 toolchain go1.23.0
 
 require (
+	github.com/Javex/caddy-fail2ban v0.0.0-20240525135356-cbf9f5a003ce
```
francislavoie, line 8: I'd rather we didn't pull in dependencies like this.

The long-term plan is to move this plugin into Caddy mainline, so we don't want any dependencies that we cannot trust to be maintained.

modules/l4fail2ban/matcher.go
trefzaxSICKAG: You're right, the condition isn't necessary.

trefzaxSICKAG:
> Hi! I actually like the idea of this matcher. As far as I understand, it requires https://github.com/Javex/caddy-fail2ban (a realtively new plugin I haven't conducted a deep analysis of, but it will be inside a go binary, so it's not a big deal then) and https://github.com/fail2ban/fail2ban (a completely separate non-go project). In other words, even if Caddy is built with caddy-l4 and this matcher included, the desired fail2ban functionality won't work unless one installs and configures fail2ban separately. Thus, it makes me doubt whether this matcher should be part of caddy-l4, https://github.com/Javex/caddy-fail2ban, or even be a separate solution.

Hi, thank you for your review!
The name of the matcher is probably a bit misleading as it basically just monitors a specific file in the file system for banned IPs. 
So you don't necessarily need to use fail2ban along with it, although that was the intention behind this module. https://github.com/Javex/caddy-fail2ban does the same for Caddy, this module provides the functionality for L4.

go.mod
trefzaxSICKAG: I have removed the external dependency and implemented the function to be part of this repository.
I would appreciate a review.

vnxme:
@trefzaxSICKAG Thanks for keeping your PR updated.

I have another topic for discussion with regards to this matcher: shall we actually name it `fail2ban`? What this matcher does is basically checking if a remote IP is present in the list stored in a predefined file. But once checked and matched, it's up to the user whether to ban it or serve anything. In my view, `fail2ban` and anything `ban`-related wouldn't suit this matcher. So I would propose `remote_ip_list` as a simple and self-explanatory alternative.

If the renaming is approved, I think the code has to be cleared to exclude any words derived from `ban` to avoid confusion.

trefzaxSICKAG:
Thank you for your suggestion @vnxme 
I have renamed the module to `remote_ip_list` and removed all variables derived from `ban`. I hope this makes the module easier to understand.

mohammed90:
At this point, this is just a `remote_ip` matcher, which already exists. Maybe the only difference is that it watches for the file update.

modules/l4remoteiplist/iplist.go
```diff
@@ -0,0 +1,176 @@
...
+		logger.Debug("could not find the file containing the IPs, trying to create it...")
+
+		// Create a new file since it does not exist
+		_, err := os.Create(ipFile)
+		if err != nil {
+			logger.Error("Error creating the IP list", zap.Error(err))
+			return nil, fmt.Errorf("cannot create a new IP list: %v", err)
+		}
+		logger.Debug("list of IP addresses successfully created")
+	}
```
vnxme, line 57: Why do you think this matcher should create the file it monitors? Wouldn't it be easier and more transparent to just load an empty list of IP addresses when the file is missing?

modules/l4remoteiplist/iplist.go
```diff
@@ -0,0 +1,176 @@
...
+	return ipList, nil
+}
+
+// Check whether a IP address is currently contained in the IP list
+func (b *IpList) IsMatched(ip string) bool {
+	// First reload the IP list if needed to ensure IPs are always up to date
+	b.reloadNeededMutex.Lock()
+	reloadNeeded := b.reloadNeeded
+	b.reloadNeeded = false
+	b.reloadNeededMutex.Unlock()
```
vnxme, line 68: I guess the unlock here makes a parallel IsMatched() check the outdated list of IP addresses. What if we unlock the mutex and change reloadNeeded to false only after a successful loadIpAddresses() return?

modules/l4remoteiplist/iplist.go
```diff
@@ -0,0 +1,176 @@
...
+		err := b.loadIpAddresses()
+		if err != nil {
+			b.logger.Error("could not load IP addresses")
+		} else {
+			b.logger.Debug("reloaded IP addresses")
+		}
+	}
+
+	for _, listIp := range b.ipAddresses {
+		if listIp == ip {
```
vnxme, line 80: You match strings, and it would probably work fine for IPv4. Consider IPv6: fd00::1 is a synonym to fd00:0:0:0:0:0:0:1, but their string representations aren't equal. So I suggest we match IP addresses, not strings.

modules/l4remoteiplist/iplist.go
```diff
@@ -0,0 +1,176 @@
...
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		lines = append(lines, scanner.Text())
+	}
+	if scanner.Err() != nil {
+		b.logger.Error("error reading the IP list", zap.Error(err))
+		return fmt.Errorf("error reading the IPs from %v", b.ipFile)
+	}
+
+	b.ipAddresses = lines
```
vnxme, line 174: I assume we have to parse IP addresses here, and filter out invalid lines (e.g. comments, spaces, etc.) by the way.

modules/l4remoteiplist/matcher.go
```diff
@@ -0,0 +1,127 @@
...
+// The Match will return true if the remote IP is found in the remote IP list
+func (m *RemoteIpList) Match(cx *layer4.Connection) (bool, error) {
+	clientIP, err := m.getRemoteIP(cx)
+	if err != nil {
+		// Error, tread IP as matched
+		m.logger.Error("error parsing the remote IP from the connection", zap.Error(err))
+		return true, err
+	}
+
+	strClientIP := clientIP.String()
```
vnxme, line 71: If we match IP addresses, no need to convert clientIP into string. Besides, `clientIP` is a bit misguiding here for me, since Caddy http app has a `client_ip` matcher, that [differs](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/ip_matchers.go) from `remote_ip` matcher. I suggest we stick to `remoteIP` to avoid confusion.

modules/l4remoteiplist/iplist.go
trefzaxSICKAG: I changed it so that it to tries to parse the IP address and throws an error if the string is not a valid IP address.
Another option would be to parse valid IP addresses and ignore errors (like comments, etc.) and maybe log a debug message (e.g. 'could not parse IP address "// comment"').
What do you think?

modules/l4remoteiplist/iplist.go
trefzaxSICKAG: This would definitely make the code cleaner. 
The problem is that at least the directory in which the IP file will later be located must exist for the file system watcher to work.

modules/l4remoteiplist/iplist.go
```diff
@@ -0,0 +1,183 @@
...
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/caddyserver/caddy/v2"
+	"github.com/fsnotify/fsnotify"
+	"go.uber.org/zap"
+)
+
+type IpList struct {
```
mholt, line 30:
```suggestion
type IPList struct {
```

modules/l4remoteiplist/iplist.go
```diff
@@ -0,0 +1,183 @@
...
+	ipFile            string         // File containing all IPs to be matched, gets continously monitored
+	ipAddresses       []netip.Addr   // List of currently loaded IP addresses that would matched
+	ctx               *caddy.Context // Caddy context, used to detect when to shut down
+	logger            *zap.Logger
+	reloadNeededMutex sync.Mutex // Mutex to ensure proper concurrent handling of reloads
+	reloadNeeded      bool       // Flag indicating whether a reload of the IPs is needed
+}
+
+// Creates a new IpList, creating the ipFile if it is not present
+func NewIpList(ipFile string, ctx *caddy.Context, logger *zap.Logger) (*IpList, error) {
```
mholt, line 40:
```suggestion
func NewIPList(ipFile string, ctx *caddy.Context, logger *zap.Logger) (*IPList, error) {
```

modules/l4remoteiplist/matcher.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/mholt/caddy-l4/layer4"
+	"go.uber.org/zap"
+)
+
+func init() {
+	caddy.RegisterModule(&RemoteIpList{})
+}
+
+type RemoteIpList struct {
```
mholt, line 33:
```suggestion
type RemoteIPList struct {
```

modules/l4remoteiplist/matcher.go
```diff
@@ -0,0 +1,126 @@
...
+	RemoteIpFile string `json:"ip_file"`
```
mholt, line 34:
```suggestion
	RemoteIPFile string `json:"ip_file"`
```

Also, it's generally a good idea for JSON struct tag names to match the names of the field name, i.e. `remote_ip_file` in this case.

mholt (changes requested):
Thanks again for the contribution/working on this. I don't really have much/any experience with fail2ban, so I will continue to defer other reviews to the others who have already commented, but I just had a couple stylistic nits. Conventionally, initialisms in Go are capitalized, so all instances of `Ip` in identifiers should become `IP`. I commented on a few of them but there are more.

I'm good with merging this, however, it does bring in fsnotify, a moderately complex dependency (not in terms of dependencies, thankfully, just code), and it would be a new dependency for Caddy if, someday, the caddy-l4 module is merged into the standard distribution. There's a good chance this would be an external plugin at that point (as would possibly some other L4 modules but I haven't looked closely yet). That's okay -- just an FYI for the future.

Thanks!

modules/l4remoteiplist/iplist.go
vnxme: If the directory doesn't exist, it's fine to return an error at provision, I think.

modules/l4remoteiplist/iplist.go
vnxme: In my view, since this file might generally be out of the user's control (e.g. created and updated by another app), it would be fine to just ignore any comments, empty lines and invalid IPs.

However, what if each line of the file can also be a network (e.g. fd00;;/8) or a range (fd00::1-fd00::f), not just a single IP?

modules/l4remoteiplist/matcher.go
vnxme: I'd vote for something simpler like `file` or `filename` as a JSON struct tag. It's the only filename field within the struct, so even the short form seems to be clear enough for me.

modules/l4remoteiplist/iplist.go
```diff
@@ -0,0 +1,183 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// Start to monitor the IP list
+func (b *IpList) StartMonitoring() {
+	go b.monitor()
+}
```
vnxme, line 89: It's more about code styling, but is there much sense in a one-line function that is called from a single place? I believe, it's just an additional execution hop we could eliminate.

modules/l4remoteiplist/iplist.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	var ipAddresses []netip.Addr
+	for _, ipStr := range lines {
+		ip, err := netip.ParseAddr(ipStr)
+		if err != nil {
+			return fmt.Errorf("invalid remote IP address: %s", ipStr)
+		}
+		ipAddresses = append(ipAddresses, ip)
+	}
```
vnxme, line 180: Is it possible to merge these two loops into one and remove the intermediate string slice?

modules/l4remoteiplist/iplist.go
vnxme: I'm probably overthinking with regards to IP ranges, but IP networks in the CIDR notation are already supported by `remote_ip`, `client_ip` and `local_ip` matchers and often included in various public blacklists.

modules/l4remoteiplist/matcher.go
trefzaxSICKAG: I changed it to `remote_ip_file`, this also matches with the module name `RemoteIPList`.

modules/l4remoteiplist/iplist.go
trefzaxSICKAG: I left it like this for the time being to simplify the code at the provision. The idea was to avoid starting goroutines there.

modules/l4remoteiplist/iplist.go
trefzaxSICKAG: I have changed it as you suggested. The provision will fail if the directory does not exist. Otherwise, if the directory exists but the file does not, no IP addresses will be loaded.

modules/l4remoteiplist/iplist.go
trefzaxSICKAG: I have ignored lines that do not represent a valid IP address.
Networks / ranges are currently not supported.

trefzaxSICKAG:
Thank you for your review @mholt 
I have changed the identifiers containing `Ip` to `IP` and renamed the JSON struct tag. Also the tests should now be successful for Windows, there was a problem with joining file paths.

modules/l4remoteiplist/iplist.go
vnxme: You can do `ip, err := caddyhttp.CIDRExpressionToPrefix(line)` instead of `ip, err := netip.ParseAddr(line)` to support networks.

trefzaxSICKAG: Thanks for the hint! Networks are supported now.

trefzaxSICKAG:
Just a quick update: I have implemented all the requested changes. From my side, this would be ready to merge.
Is there anything else I should update?

modules/l4remoteiplist/iplist.go
```diff
@@ -0,0 +1,186 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// Start to monitor the IP list
+func (b *IPList) StartMonitoring() {
+	go b.monitor()
+}
```
mohammed90, line 84: Just to double check, this won't leak goroutines on every reload because Caddy closes/shuts the `caddy.Context` on config reload and/or shutdown, right?

This is the only concern I have.

modules/l4remoteiplist/matcher.go
```diff
@@ -0,0 +1,126 @@
...
+		ID:  "layer4.matchers.remote_ip_list",
+		New: func() caddy.Module { return new(RemoteIPList) },
+	}
+}
+
+// Provision implements caddy.Provisioner.
+func (m *RemoteIPList) Provision(ctx caddy.Context) error {
+	m.logger = ctx.Logger()
+
+	remoteIPList, err := NewIPList(m.RemoteIPFile, &ctx, m.logger)
```
mholt, line 52:
```suggestion
	remoteIPList, err := NewIPList(m.RemoteIPFile, ctx, m.logger)
```

Passing a pointer to a context is probably a bug...

modules/l4remoteiplist/iplist.go
mholt: Good eye. It does seem to check the context in its loop, and return after an error.

trefzaxSICKAG: As @mholt said, the main loop within the goroutine currently returns as soon as `ctx.Done()` is true or an error occurs.
Is there anything else that should be checked to prevent a goroutine leak? Is `ctx.Done()` always true when the configuration is reloaded?

trefzaxSICKAG: @mohammed90 @mholt I tested the behaviour during a reload of the Caddy config: the context gets closed by Caddy and the goroutine stops because ctx.Done() is true.
To make sure the goroutine does not leak, I also implemented the caddy.CleanerUpper interface in the matcher as a fallback, which then stops the goroutine via a separate channel.

trefzaxSICKAG:
Any updates on this?
From my side the implementation would be complete at the moment.

Short summary of the features:
 - new matcher `remote_ip_list` which monitors a file with a list of IP addresses and/or IP ranges and matches incoming requests if the client IP is included in the list
 - can be used alone or in conjunction with fail2ban to monitor log files and ban certain clients without having to manipulate the host system's iptables, enabling e.g. rootless docker setups with caddy + fail2ban
   - can be used to protect protocols on layer 4 against brute force attacks (FTP, SSH, OPC UA, SNMP, IMAP, POP3, SMTP, LDAP, MQTT...)
 - now also implements caddy.CleanerUpper to avoid goroutine leaks

mholt:
I think we can give this a go. Any other comments, @mohammed90 ?

mohammed90:
> Any other comments, @mohammed90 ?

Nope, nothing else. LGTM.

Issue #274, opened 2024-12-06
3 second timeout for requests when using listener_wrappers `tls` sending requests to `reverse_proxy`


Monviech:
Downstream report: https://github.com/opnsense/plugins/issues/4384

The issue is, that when something is going through the layer4 `listener_wrappers` to the `tls directive` into the `reverse proxy`, there is a timeout of 3 seconds after which the context gets canceled.

- It does NOT happen with only the "reverse_proxy" proxying the request
- It does NOT happen with only the layer4 modules "proxy" proxying the request

As example, there is an API endpoint in OPNsense that could take anywhere from 10 to 30 seconds to respond, depending on different factors. After exactly 3 seconds this context gets canceled by the client (Caddy) using this Caddyfile:

```
{
	servers {
		protocols h1 h2 h3
		listener_wrappers {
			layer4 {
			}
			tls
		}
	}
}

:80 {
}
:443 {
}

opn.example.com {
	handle {
		reverse_proxy https://172.16.0.254:4444 {
			transport http {
				tls_insecure_skip_verify
			}
		}
	}
}
```

Relevant error can be seen in downstream report.
I have tested this and can reproduce it.

mholt:
Hmm I'm wondering if there's a timer of 3 sec set anywhere in the code. Can look more into it soon!

Monviech:
The only default I found with exactly 3 seconds was `dial_timeout` but that was a caddyserver directive.

mohammed90:
It's this
https://github.com/mholt/caddy-l4/blob/3c6cc2c0ee0875899fde271fbdef95be3fef7a92/layer4/server.go#L33

which is injected here
https://github.com/mholt/caddy-l4/blob/3c6cc2c0ee0875899fde271fbdef95be3fef7a92/layer4/listener.go#L57


and set here:
https://github.com/mholt/caddy-l4/blob/3c6cc2c0ee0875899fde271fbdef95be3fef7a92/layer4/routes.go#L122-L123

mholt:
Good find. Yeah, so it sounds like after 3 seconds, matchers are being cancelled. I wonder if that's because there are no matchers configured?

mohammed90:
> matchers are being cancelled. I wonder if that's because there are no matchers configured?

My understanding is the timeout of the matchers is wrongly passed to the connection read deadline. The timeout deadline isn't updated after matching.

mohammed90:
I take it back. It's reset here.

[...]

PR #282, opened 2025-01-02
remove deadline when fallback route is chosen

WeidiDeng:
Fix [274](https://github.com/mholt/caddy-l4/issues/274).


mholt (approved):
Let's give it a shot! Sounds like it solves the problem being addressed. Thank you, Weidi, for the patch!!

PR #289, opened 2025-01-23
proxy: support dynamic port number

lelemka0:
This PR delays address parsing in some case to make dynamic port numbers work properly. 
If dial addresses contains `{` and `}`, address parsing in `Upstream.provision` will be skipped then done later in `Handler.dialPeers` after all placeholders are replaced, and active health checks will be ignored.

Example Caddyfile:
```
layer4 {
	:2000 {
		@test tls sni_regexp ^([0-9]{1})\.test$
		route @test {
			proxy {
				upstream 127.0.0.1:800{tls.regexp.1}
			}
		}
	}
}
```


modules/l4proxy/upstream.go
```diff
@@ -46,6 +57,12 @@ type Upstream struct {
 	// have before being marked as unhealthy (if > 0).
 	MaxConnections int `json:"max_connections,omitempty"`
 
+	// Is dial a dynamic upstream
+	// If true, skip address parsing in Upstream.provision
+	// then parse the address after replacing all placeholders in Handler.dialPeers.
+	// (support dynamic port)
+	Dynamic bool `json:"dynamic,omitempty"`
```
mohammed90, line 64: No need for a new field. I think it's better that the need for delayed parsing is inferred if the dial address contains `{` and `}`. Pretty sure that's what we did in Caddy `reverse_proxy` handler.

lelemka0: The condition has been changed to the address contains `{` and `}` and the new field has been removed. Thank you.

lelemka0:
Will this be merged, or will other plans be considered for dynamic ports?

modules/l4proxy/upstream.go
```diff
@@ -30,12 +30,23 @@ import (
 	"github.com/mholt/caddy-l4/layer4"
 )
 
+func ParseAddress(addr string) (*caddy.NetworkAddress, error) {
```
mholt, line 33: Can we add a godoc to this? to explain why it is needed? (I'm not entirely sure myself, just by looking at it, since it uses only exported methods)

mholt (changes requested):
Sorry, been super busy... I haven't tested this myself but I trust you have? I just have one request for a comment, as I'm not sure if that new exported function is necessary.

lelemka0:
This was my mistake about the exported function, it's not used outside and is unlikely to be used outside.
I have modified it to unexported and added doc.

I have been using it and it is working fine for two weeks. For a simple test, use the following config:
```
{
	debug
	cert_issuer internal
	layer4 {
		:2000 {
			@test tls sni_regexp ^(\d{1})\.test$
			route @test {
				proxy {
					upstream 127.0.0.1:820{tls.regexp.1}
				}
			}
		}
	}
}

:8200 {
	tls {
		on_demand
	}
	respond OK 200
}
:8201 {
	tls {
		on_demand
	}
	respond OK-1 200
}
```
```
curl --insecure --resolve "1.test:2000:127.0.0.1" "https://1.test:2000/"
```

modules/l4proxy/upstream.go
vnxme: It has been fixed by the PR author. The function is actually needed to reuse the address parsing code between `proxy.go` and `upstream.go`.

vnxme:
@lelemka0, thank you for contributing! I've merged the current master into your branch and made a few minor adjustments. I can also confirm the test you suggested above works for me.

@mholt, is there anything else you would like to change in this PR before it may be merged?

mholt (approved):
Nah, this is looking good. Thanks!

bermeitinger-b:
Please note that this breaks (at least) the DoT example with being unable to find host "tcp/172.16.0.1:53":

```
        layer4 {     
                :853 {                          
                        @dot tls sni <myhostname>
                        route @dot {                                                            
                                tls                                                             
                                proxy tcp/172.16.0.100:53
                        }                                                                       
                }
        }
```

The IP address is my pihole instance at that port. Only this commit breaks the setup.

vnxme:
@bermeitinger-b, thank you for testing and reporting the issue. Please check if it works with #409.

PR #297, opened 2025-02-13
Add example for how to combine Caddy’s regular HTTP app and L4

ntninja:
Fortunately, the listener_wrappers syntax works exactly how you would expect for adding L4 rules before regular HTTP handling!


mholt (approved):
Thanks for this!

Actually, it was discussed recently that maybe we should move all the examples into the Wiki.

How would you feel if we started now? Is that something you'd be willing to help us with?

Issue #269, opened 2024-11-17
UDP Proxy Protocol not functioning


erroltuparker:
Hi All,

It appears that there is some issue with the proxy protocol implemented for UDP as my server recieves this response when trying to proxy.

```
[2024-11-18 00:00:04 Local] [192.168.200.1:37825] [UDPPROXY] System.ArgumentNullException: Value cannot be null. (Parameter 'address')
   at System.ArgumentNullException.Throw(String paramName)
   at DnsServerCore.Dns.DnsServer.ReadUdpRequestAsync(Socket udpListener, DnsTransportProtocol protocol)
[2024-11-18 00:00:04 Local] [192.168.200.1:37825] [UDPPROXY] System.IO.InvalidDataException: The stream does not contain PROXY protocol header.
   at TechnitiumLibrary.Net.ProxyProtocol.ProxyProtocolStream.CreateAsServerAsync(Stream baseStream, CancellationToken cancellationToken) in Z:\Technitium\Projects\TechnitiumLibrary\TechnitiumLibrary.Net\ProxyProtocol\ProxyProtocolStream.cs:line 138
   at DnsServerCore.Dns.DnsServer.ReadUdpRequestAsync(Socket udpListener, DnsTransportProtocol protocol) in Z:\Technitium\Projects\DnsServer\DnsServerCore\Dns\DnsServer.cs:line 409
```

But when I use nginx, it works fine. I did raise this with the provider and they also confirmed my findings

https://github.com/TechnitiumSoftware/DnsServer/issues/1107

TCP works fine for proxying as I'm also utilizing that, but I would like to cover both sides of it.

The thing I see on the caddy side when this is run is the below errors.

```
{"level":"debug","ts":1731887445.6253898,"logger":"layer4","msg":"matching","remote":"172.253.247.61:39349","error":"consumed all prefetched bytes","matcher":"layer4.matchers.dns","matched":false}
{"level":"debug","ts":1731887445.6254132,"logger":"layer4","msg":"prefetched","remote":"172.253.247.61:39349","bytes":48}
{"level":"debug","ts":1731887445.6254213,"logger":"layer4","msg":"matching","remote":"172.253.247.61:39349","matcher":"layer4.matchers.dns","matched":true}
{"level":"debug","ts":1731887446.644327,"logger":"layer4","msg":"matching","remote":"172.68.209.70:19458","error":"consumed all prefetched bytes","matcher":"layer4.matchers.dns","matched":false}
[...]

PR #305, opened 2025-05-21
only enable proxy protocol if the network is not unix or if the netwo…

WeidiDeng:
…rk is unixgram

fix [issue here](https://github.com/mholt/caddy-l4/issues/269#issuecomment-2602461506)


mholt (approved):
Thanks -- I've been so behind from being sick and I'll be catching up for a while.

PR #307, opened 2025-05-21
extract local address for udp sockets on windows and linux

WeidiDeng:


vnxme:
@WeidiDeng, could you please comment on the purpose of these changes? Do they allow for logging UDP connections properly? Was this PR intended to be merged after #305?

WeidiDeng:
For udp sockets, by default, operating system only provides remote address for every packet read, not local address, i.e., the address in which the packet comes from. To acquire this address, special option needs to be turned on for the socket, and we will use another method to read the packet and its associated address data. This address data has a format that's platform dependent and will be parsed as the platform requires.

It enhances the proxy protocol for udp, which for now forwards the dummy address `127.0.0.1` as the local address if not available because stdlib will return the address we pass to `ListenPacket` and most of the time it's `0.0.0.0` and invalid for proxy protocol.

I guess no one cares the local address as long as the proxy protocol works for their application.

PR #308, opened 2025-05-30
proxy: make a few methods public

matdibu:
This PR renames a few methods in `l4proxy`:

- `LoadBalancing.tryAgain` to `TryAgain`
- `Handler.proxy` to `Proxy`
- `Handler.dialPeers` to `DialPeers`

I needed this to implement a custom `l4proxy` with a modified `DialPeers`

Hopefully, the changes are not too intrusive :)


mohammed90:
What are you trying to do? The analogous methods in core Caddy are also unexported, so it's hard to find their status here is a blocker for extensions.

matdibu:
A post-tls handshake using ExportKeyingMaterial.

I've tried it and it works, chaining the handler after `tls`.

But I couldn't add it to the l4proxy module without changing a few things because I have to run it right after the tls handshake.

Do you know a better way to achieve this?

mholt:
Exporting these methods is a bit weird, in that it implies the Handler and LoadBalancing types is being used in a way that is only intended for the l4proxy package to implement.

Do you feel like your goals would be better accomplished with some sort of hook into the proxy?

mholt:
Is the key material of the _upstream_ needed? If so I could see why the proxy might be a good place for it since it connects to the upstream.

matdibu:
> Do you feel like your goals would be better accomplished with some sort of hook into the proxy?

sure; the end goal is just to execute a handshake procedure after TLS is established

executing the server's handshake was simple enough, I just added a handler after the TLS handler

but for the client, I needed to "inject" it immediately after TLS was established, which I couldn't figure out how to do aside from modifying l4proxy directly -- is/was there a simpler alternative? I only have one upstream

> Is the key material of the upstream needed?

exactly

here's how I implemented the `customtls` handler
```go
func (t *Handler) Handle(cx *layer4.Connection, next layer4.Handler) error {
	tlsConn := cx.Conn.(*tls.Conn)

	err := customtls.Server(tlsConn).Handshake()
	if err != nil {
		return err
	}

	return next.Handle(cx.Wrap(tlsConn))
}
```

and the relevant bits in `l4customproxy`
```go
type Handler struct {
	l4proxy.Handler

	ctx    caddy.Context
}

func (h *Handler) DialPeers(upstream *l4proxy.Upstream, repl *caddy.Replacer, down *layer4.Connection) ([]net.Conn, error) {
	peers, err := h.Handler.DialPeers(upstream, repl, down)
	if err != nil {
		return nil, err
	}

	for _, peer := range peers {
		tlsConn, ok := peer.(*tls.Conn)
		if !ok {
			return nil, fmt.Errorf("tried to perform customTLS handshake, but TLS is disabled")
		}

		err = customtls.Client(tlsConn).Handshake()
		if err != nil {
			return nil, err
		}
	}

	return peers, nil
}
```

where both `customtls.Client(tlsConn).Handshake()` and `customtls.Server(tlsConn).Handshake()` do
```go
cs := c.ConnectionState()

cs.ExportKeyingMaterial(...)
```

mholt:
I see, you're embedding the Handler type and then writing your own methods that get promoted over the embedded one's, but you still call the embedded Handler's method and are basically wrapping it.

I dunno. I guess that's fine with me. The alternative is to make the Dialer a module that gets loaded dynamically. So you can do the dialing in a custom way.

mholt:
Ah, sorry, I got busy with the rest of my backlog.

PR #312, opened 2025-06-30
bump: github.com/quic-go/quic-go to v0.53.0

IceCodeNew:
fix #311

PR #315, opened 2025-08-06
l4quic: Close quic.EarlyListener only if it has accepted a new connection

vnxme:
This change fixes a bug and is required before we upgrade dependencies to `go` v.1.24, `caddyserver/caddy` v2.10.0 and `quic-go` v0.51.0. See also https://github.com/mholt/caddy-l4/pull/312#issuecomment-3160302145.


mholt (approved):
Thanks @vnxme for picking up these related problems! Will start by merging this and then we can tidy up the dependencies/test failures.

PR #316, opened 2025-08-06
Upgrade dependencies: caddyserver/caddy v2.10.0 + quic-go v0.54.0

vnxme:
This PR only upgrades `caddyserver/caddy` to v2.10.0 (+ `go mod tidy`), but it is based on the `l4quic` matcher fix to ensure all the tests pass. I will rebase it once we merge #315.


francislavoie:
@vnxme what that would do is it make anyone who builds with caddy-l4 to get forced onto that version of Caddy instead of the one they actually requested (e.g. if they build `xcaddy build 2.10.0 --with github.com/mholt/caddy-l4` they would instead get whatever commit from Caddy you point caddy-l4 to). But honestly, that's fine cause caddy-l4 is still essentially experimental.

vnxme:
@francislavoie I don’t insist on bumping to a pseudo-version here and would be perfectly fine with v2.10.0. Then we would have to ignore automatic tests failures with Caddy master and users’ complaints about inability to build any commit of Caddy after quic-go upgrade with this app. Though I would rather release Caddy v2.10.1 instead to take into account the code breaking change in quic-go.

mholt (approved):
Thank you very much, again!

I updated it with #315 -- shall we just merge this or was there more you wanted to change?

vnxme:
@IceCodeNew, as far as fsnotify and things-go/go-socks5 upgrades are concerned, they aren't critical at all. However, my intention was to let the users benefit from the bug fixes available in the upstreams.

Monviech:
@vnxme Thank you :)

PR #320, opened 2025-08-14
refactor: Minor fixes

vnxme:
This PR includes a number of minor code style fixes and, *inter alia*, closes #310.

PR #321, opened 2025-08-14
refactor: Adjust dependabot and lint workflows, fix linter issues

vnxme:
This PR deals with the dependabot and lint workflows based on what we have in the mainline Caddy.


mholt:
Thanks! Looks like it will need some polish before the CI checks all pass.

Really appreciate it :)

vnxme:
@mholt, I suggest we merge #320, then I'll rebase this PR and finally fix the remaining linter warnings.

mohammed90:
I recommend adding the `.golangci.yml` to configure and control the lint rules of golangci-lint. Otherwise it'll run with its own rules and default config.

https://github.com/caddyserver/caddy/blob/master/.golangci.yml

mholt:
Yeah, let's remove all deprecated stuff in Caddy from this repo, since they'll eventually be removed from Caddy too.

Sorry for the delay on this -- once this is synced up I can merge it in!

vnxme:
Almost green. I don't understand what's wrong with the linter tests on Windows.

mohammed90:
I just copied what we have in caddy repo. We can certainly make it the same for all files by using the pattern `*` without `.go`.

vnxme:
OK, then I'm fine with keeping .gitattributes in sync with Caddy.

Is there anything else we should do before merging?

PR #344, opened 2025-11-18
Docs

vnxme:
### Summary

This is neither yet another *please write some docs* request, nor a limited set of configuration examples. This PR contains a complete set of files describing all the matchers and handlers included in the package, as well as routes, servers and the overall architecture.

### Description

I know, nobody likes writing docs. Many enjoy reading them though, as they save the time spent on unsuccessful configuration attempts. I can see from a number of issues that Caddy-L4 users tend to ask basic questions again and again because they have no reliable source of information on how to configure this powerful app, unless they can read and understand the code. Hope, now we will be able to provide references to the docs to help them. I expect the docs will contribute to the project growth.

I've done my best to describe everything: all the supported fields/options, placeholders support, etc. Nevertheless, I suppose something might have been missed. Anyone is welcome to make additions and/or corrections they consider valuable for others.

The documentation files are placed in the docs directory with matchers, handlers and examples grouped into corresponding sub-directories. The templates sub-directory contains template files to be used when adding new examples, matchers or handlers.

A few integration tests have been changed to sync with the docs. I have also updated the readme to reference the docs, but I think it may require a complete rewrite with the index page of the docs further updated.

### Further steps

I suggest we could close some issues if they are covered by the docs once we merge this PR.

### DIFF
--- a/docs/matchers/regexp.md
+++ b/docs/matchers/regexp.md
@@ -14,7 +14,7 @@ The matcher provides `pattern` field that contains a regular expression to match
 supports [placeholders](https://caddyserver.com/docs/conventions#placeholders) which are resolved at provision.
 
 The matcher also has `count` field that sets the number of bytes read from the beginning of the first packet
-to match against. By default, it equals `4`. It shouldn't exceed `8 KiB` (`MaxMatchingBytes` constant value) which is
+to match against. By default, it equals `4`. It shouldn't exceed `16 KiB` (`MaxMatchingBytes` constant value) which is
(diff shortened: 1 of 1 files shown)


felixauringer:
Sorry for reviving this.
I just want to say that I am heavily using caddy-l4 in my home network and it always took some time to configure it because I had to read through all the method comments (which is fair for an experimental project).
But those days are over and it's now easier to use than ever!
Thank you so much for contributing this @vnxme :heart:

PR #346, opened 2025-11-24
Support fragmented ClientHello and increase matching buffer

cdn0x12:
This pr addresses issues with TLS ClientHello messages that are fragmented across multiple TLS records, which is increasingly common with larger handshakes involving post-quantum cryptography extensions like ML-KEM. Previously, the TLS matcher assumed the entire ClientHello would be present in the first read, causing SNI parsing failures when the handshake message was split. The updated logic now correctly reassembles the ClientHello by reading subsequent records until the full handshake message is obtained, ensuring SNI extraction under aggressive fragmentation.

Additionally, the maximum buffer size for matching (`MaxMatchingBytes`) has been increased from 8KB to 16KB. The previous limit might be insufficient for TLS handshakes that may include large post-quantum key shares, extensive session tickets, or padding. This increase provides more headroom to support these larger messages.

### DIFF
--- a/modules/l4tls/matcher.go
+++ b/modules/l4tls/matcher.go
@@ -108,6 +108,7 @@ func (m *MatchTLS) Match(cx *layer4.Connection) (bool, error) {
 			break
 		}
 
+		// nolint:gosec // disable G602 // https://github.com/securego/gosec/issues/1406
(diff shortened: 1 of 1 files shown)


mholt:
Thank you for the patch!

Are the lint warnings valid here, or false positives? (Maybe we need some length checks to confirm it won't crash? Not sure.)

PR #347, opened 2025-11-25
DNS handler [WIP]

vnxme:
### Summary

This PR contains a DNS handler implementation based on [miekg/dns](https://github.com/miekg/dns). Caddy requires this package internally, so there are no extra dependencies (yet). The handler supports both *passthrough* and *smart* handling schemes: the former runs `miekg/dns.Server` under the hood, while the latter provides a lightweight alternative, which is preferred by default.

### Design

The handler follows a modular design inspired by Caddy -- it supports extending caches, zones, upstreams and providers.
- **Caches** make the handler more responsive and save traffic: `simple` cache is a primitive independent in-memory key-value storage with virtually no optimizations, while `shared` cache allows to sync caches of multiple DNS handlers (e.g., think of listening on both TCP and UDP) and save memory by using a single underlying cache. More caches could be implemented, e.g. `file`, `redis`, etc.
- **Zones** control how the handler reacts to incoming requests: looks up in its own memory (`auth` which stands for *authoritative*), communicates to its upstreams (`rec` which stands for *recursive*) or sends back the client's IP address mainly for debug purposes (`ref` which stands for *reflective*); but more zones could be implemented as well.
- **Providers** represent sources of DNS records for authoritative zones, with `text` being the only one included in the package. More providers could be implemented, e.g. `file`, `sql`, etc.
- **Upstreams** are channels of DNS records sourcing for recursive zones:  `udp` makes unencrypted UDP-only requests, `tcp` makes unencrypted TCP-only requests, and `plain` makes unencrypted UDP requests followed by TCP requests, if needed. More providers could be implemented, e.g. `doh`, `dot`, `doq`, etc.

### Disclaimer

My intention was to make this handler a really powerful addition on Caddy to let this software replace Bind, PowerDNS, Unbound, etc., especially on small-scale deployments. I don't think performance should be a comparison criterion -- it's the ability to run everything (DNS, HTTPS, etc.) in one binary and configure everything with one caddyfile that makes this solution amazing for me, together with unlimited extensibility.

Unfortunately, this is still a WIP, i.e. neither has the handler been thoroughly tested by me, nor have I managed to finalise the solution and implement all the features I wanted. However, it's fully functional at this stage. I've published it here in case there is anyone who wants to try it out and/or contribute to its development.

### Configuration

Below is a part of my caddyfile used for development and testing. For now, I put it here with no comments -- the relevant explanations could be easily found in the code (e.g. see comments to `UnmarshalCaddyfile` functions), or I may provide them later.

```caddyfile
{
    layer4 {
        udp/:5301 {
            route {
                dns {
                    #cache shared
                    #cache simple
                    #compress
                    zone auth caddy. {
                        expand_cname
                        provider text {
                            records {
                                caddy. IN SOA ns.caddy. mail.caddy. 12345 86400 7200 604800 172800
                                abc.caddy. IN A 127.0.0.1
                                abc.caddy. 3600 IN AAAA ::1
                                $TTL 86400
                                abc IN MX 10 mail
                                $ORIGIN def.caddy.
                                test.caddy. IN CNAME test
                                test IN CNAME abc.caddy.
                                $GENERATE 0-3 srv-$ A 127.0.1.$
                                * IN CNAME xyz.caddy.
                                xyz.caddy. IN A 127.0.0.2
                                xyz.caddy. IN AAAA ::2
                                dn1.caddy. IN DNAME dn3.caddy.
                                dn3.caddy. IN DNAME dn2.caddy.
                                dn2.caddy. IN A 127.0.0.3
                                test1.dn2.caddy. IN A 127.0.0.4
                                test2.dn2.caddy. IN CNAME test.caddy.
                            }
                        }
                        support_wildcard
                    }
                    zone rec example.com. {
                        cache shared
                        #cache simple

                        #upstream plain [::1]:5302
                        upstream plain 1.0.0.1:53 {
                            prefer_tcp
                        }
                        #upstreams_random
                        #upstreams_sequential
                    }
                    zone ref whoami.caddy.
                }
            }
        }

        udp/:5302 {
            route {
                dns {
                    zone auth example.com. {
                        provider text {
                            records {
                                example.com. IN A 127.0.0.2
                                example.com. IN AAAA ::2
                            }
                        }
                    }
                }
            }
        }

        tcp/:5301 {
            route {
                dns {
                    #cache shared
                    #cache simple
                    compress

                    zone auth caddy. {
                        expand_cname
                        provider text {

                            records {
                                caddy. IN SOA ns.caddy. mail.caddy. 12345 86400 7200 604800 172800
                                abc.caddy. IN A 127.0.0.1
                                abc.caddy. 3600 IN AAAA ::1
                                $TTL 86400
                                abc IN MX 10 mail
                                $ORIGIN def.caddy.
                                test.caddy. IN CNAME test
                                test IN CNAME abc.caddy.
                                $GENERATE 0-3 srv-$ A 127.0.1.$
                                * IN CNAME xyz.caddy.
                                xyz.caddy. IN A 127.0.0.2
                                xyz.caddy. IN AAAA ::2
                                dn1.caddy. IN DNAME dn3.caddy.
                                dn3.caddy. IN DNAME dn2.caddy.
                                dn2.caddy. IN A 127.0.0.3
                                test1.dn2.caddy. IN A 127.0.0.4
                                test2.dn2.caddy. IN CNAME test.caddy.
                            }
                        }
                        support_wildcard
                    }
                    zone rec example.com. {
                        cache shared
                        #cache simple

                        #upstream plain [::1]:5302
                        upstream plain 1.0.0.1:53 {
                            prefer_tcp
                        }
                        #upstreams_random
                        #upstreams_sequential
                    }
                    zone ref whoami.caddy.
                }
            }
        }
    }
}
```

### Steps before merging
- [ ] Compose automatic tests for the handler and all its modules
- [ ] Provide caddyfile integration tests for the handler
- [ ] Describe structures and options in `docs/handlers/dns.md`
- [ ] Mention the handler in `docs/handlers.md` and/or `README.md`

### DIFF
--- a/modules/l4dns/caches.go
+++ b/modules/l4dns/caches.go
@@ -163,21 +163,18 @@ func (c *CacheSimple) Get(r *dns.Msg) (*dns.Msg, error) {
 }
 
 // Provision spawns a goroutine to clear c's internal store.
-func (c *CacheSimple) Provision(ctx caddy.Context) error {
+func (c *CacheSimple) Provision(_ caddy.Context) error {
 	go func() {
 		t := time.NewTicker(DefCacheSimpleClearInterval)
-		for {
-			select {
-			case <-t.C:
-				t.Stop()
-				c.store.Range(func(k, v any) bool {
-					if !v.(*CacheSimpleStoreValue).IsValid() {
-						c.store.Delete(k)
-					}
-					return true
-				})
-				t.Reset(DefCacheSimpleClearInterval)
-			}
+		for range t.C {
+			t.Stop()
+			c.store.Range(func(k, v any) bool {
+				if !v.(*CacheSimpleStoreValue).IsValid() {
+					c.store.Delete(k)
+				}
+				return true
+			})
+			t.Reset(DefCacheSimpleClearInterval)
 		}
 	}()
 	return nil
--- a/modules/l4dns/dispatcher.go
+++ b/modules/l4dns/dispatcher.go
@@ -223,7 +223,7 @@ func (i *Inbox) Validate(secret string, prev *dns.TSIG) (*dns.TSIG, error) {
 		return nil, ErrInboxValidateNoSecret
 	}
 	if prev != nil {
-		return sig, dns.TsigVerify(i.bytes, secret, prev.MAC, !(len(prev.MAC) == 0))
+		return sig, dns.TsigVerify(i.bytes, secret, prev.MAC, len(prev.MAC) != 0)
 	}
 	return sig, dns.TsigVerify(i.bytes, secret, "", false)
 }
@@ -280,7 +280,7 @@ func (o *Outbox) PushSign(msg *dns.Msg, secret string, prev *dns.TSIG) (err erro
 	o.msg = msg
 	o.msg.Compress = o.compress
 	o.msg.SetTsig(prev.Hdr.Name, prev.Algorithm, prev.Fudge, time.Now().Unix())
-	o.bytes, _, err = dns.TsigGenerate(o.msg, secret, prev.MAC, !(len(prev.MAC) == 0))
+	o.bytes, _, err = dns.TsigGenerate(o.msg, secret, prev.MAC, len(prev.MAC) != 0)
 	if err != nil {
 		return err
 	}
--- a/modules/l4dns/io.go
+++ b/modules/l4dns/io.go
@@ -52,7 +52,7 @@ func ReadBytesFromTCP(cx *layer4.Connection) ([]byte, error) {
 
 	// Read a minimum number of bytes
 	buf := make([]byte, length+1)
-	n, err := io.ReadFull(cx, buf[:dnsHeaderBytes])
+	_, err := io.ReadFull(cx, buf[:dnsHeaderBytes])
 	if err != nil {
 		return nil, err
 	}
@@ -63,6 +63,7 @@ func ReadBytesFromTCP(cx *layer4.Connection) ([]byte, error) {
 	}
 
 	// Read the remaining bytes
+	var n int
 	n, err = io.ReadAtLeast(cx, buf[dnsHeaderBytes:], int(length-dnsHeaderBytes))
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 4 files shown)


mholt:
This is actually pretty freakin' cool. It's like CoreDNS (?) -- which is based on Caddy 1 -- but for Caddy 2.

Anyway, I very much like where this is going.

vnxme:
@Monviech What do you mean by “injecting policy decisions into the proxied connections” and “infusing policy decisions into higher layer protocols”? I haven’t caught the idea of your comment above.

Monviech:
I only mean it like that changing the content of DNS packets or adding new options to them, is related to policy decisions.
Should the proxy become a source of truth for this?

Right now, if I see the scope of the layer 4 proxy correctly, it should just match and forward based on patterns, not change the content of these packets per se.

I think this is cool and innovative, but am also cautious because it's unsure where the road will lead to. I'm just curious here.

vnxme:
@Monviech OK, thanks for clarification. This is a *handler*, while you’re talking about a *matcher*. These are completely separate features, even though they are both related to DNS.

The `dns` **matcher** lets Caddy route traffic if it resembles DNS and, if applicable, has some inner options, flags, etc. (e.g. DNS requests for abc.com vs DNS requests for def.com vs TLS traffic). It is often used in combination with other matchers to serve multi-protocol traffic on a single port. What’s important to note is a matcher never changes packets or sends anything back.

The `dns` **handler** lets Caddy handle DNS requests in a completely independent way, i.e. makes it a DNS authoritative server, recursor and/or whatever you can imagine and implement. It expands Caddy’s capabilities and makes this software even more universal and versatile. For example, you may want to serve DNS requests for abc.com on your own while proxying DNS requests for def.com to an upstream — the self-serving part of this example is what you may need the handler for. A handler is never used for traffic matching.

Monviech:
Ah I see, so I assume the main benefit is convenience.

Though if one would follow this reasoning, any protocol could be essentially compiled as handler into caddy + caddy-l4. E.g. a DHCP/DHCPv6+RA implementation. And then it would be like reinventing dnsmasq inside caddy.

My only concern is that its currently unclear where the line could be drawn in terms of scope.

In no way my intention is to discredit any effort in any way. My sole intention was to crystalize this realisation out of our discussion.

PR #360, opened 2025-12-20
proxyprotocol: use `pires/go-proxyproto` module

### LINKED ISSUE #176: Switch to `go-proxyproto`
Caddy 2.8 has [changed their PROXY protocol dependency to `go-proxyproto`](https://github.com/caddyserver/caddy/pull/5915) which is a bit more featureful (_a port can support PROXY protocol connections but optionally still accept connections without PROXY header present based on a configured policy, where the PROXY protocol header is only accepted from trusted hosts_).

`caddy-l4` should probably follow that change?

---

Apparently making the switch will [require a big refactor](https://github.com/mholt/caddy-l4/issues/269#issuecomment-2892326307)? There's a known issue with UDP support currently, unclear if that improves when migrating to `pires/go-proxyproto`

mohammed90:
Copilot did the initial run. The `proxy.go` file didn't need any changes by me. The `handler.go` file required manual intervention, but most of the necessary changes were copied from Caddy repo. The file `policy.go` is copied from Caddy repo as-is.

As a bonus, this adds `deny` list as well as `fallback_policy` .

Closes #176

### DIFF
--- a/layer4/routes.go
+++ b/layer4/routes.go
@@ -111,9 +111,11 @@ func (routes RouteList) Compile(logger *zap.Logger, matchingTimeout time.Duratio
 
 		var (
 			lastMatchedRouteIdx = -1
-			lastNeedsMoreIdx    = -1
-			routesStatus        = make(map[int]int)
-			matcherNeedMore     bool
+			// it actually records where a matcher may require more data, i.e. the one that really needs more,
+			// the one after a matched route, the one after a not matched route with no previous routes needing more.
+			// -1 means there is no route that requires more data
+			lastNeedsMoreIdx = -1
+			routesStatus     = make(map[int]int)
 		)
 		// this loop should only be done if there are matchers that can't determine the match,
 		// i.e. some of the matchers returned false, ErrConsumedAllPrefetchedBytes. The index which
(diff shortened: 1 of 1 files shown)


vnxme:
@mohammed90, thank you! It looks good to me.

In addition to the minor question above, I suggest we also adjust [integration/caddyfile_adapt/gd_handler_proxyprotocol.caddytest](https://github.com/mholt/caddy-l4/blob/master/integration/caddyfile_adapt/gd_handler_proxyprotocol.caddytest) to cover the extended Caddyfile syntax (`deny` and `fallback_policy` options).

vnxme (approved):

modules/l4proxyprotocol/handler.go
```diff
@@ -41,8 +41,46 @@ type Handler struct {
...
+	//   a PROXY header is not present, subsequent reads do not. It is the task
+	//   of the code using the connection to handle that case properly.
+	//
+	// - SKIP: accepts a connection without requiring the PROXY header.
+	//   Note: an example usage can be found in the SkipProxyHeaderForCIDR
+	//   function.
+	//
+	// Default: IGNORE
+	//
+	// Policy definitions are here: https://pkg.go.dev/github.com/pires/go-proxyproto@v0.7.0#Policy
```
mohammed90:
```suggestion
	// Policy definitions are here: https://pkg.go.dev/github.com/pires/go-proxyproto#Policy
```

PR #361, opened 2025-12-23
Support upstream source addresses and DNS resolver preferences

JDTX0:
This adds support for specifying the source address used for upstreams and adds functionality to control resolver preferences.

## Source address features
An IPv4 and/or IPv6 address can be specified. It's possible to specify both at once and it will automatically choose the correct source address based on the resolved address of the upstream is it connecting to.

For example, if we're dialing an IPv6 address, it will choose the IPv6 source address specified in `local_address`, likewise for IPv4, it will look for an IPv4 address in `local_address` and so on.

This is useful in cases where we want to control what the upstream endpoint sees for our source IP and handles cases where the upstreams are dual-stack (supporting v4 and v6)

If an upstream is dual-stack and you've only specified a single address family in local_adddress (i.e. v4 or v6 only), then it will dial using default options (kernel-chosen address) unless the resolver preference is restricted to a specific family.

`local_address` is a list of strings in JSON and space-separated (or multi-line, same as `dial`) in a Caddyfile, supports placeholders (known ones replaced at provision, the rest per-connection), and is not supported on Unix socket upstreams (rejected at provision).

## Resolver preferences
This also adds a new setting called `resolver_preference` scoped to the `upstreams` section. This allows you to control what address is dialed based on your preference.

For example:
- `ipv4_only` - Even if the resolved hostname has AAAA (IPv6) records, we'll only dial out to the IPv4 addresses
- `ipv6_only` - Same as above but reversed, but only dials out to IPv6 addresses.
- `ipv4_first` - If a domain has both A/AAAA records, prefer the A (IPv4) record to dial out to.
- `ipv6_first` - Same as above, but prefer AAAA (IPv6) addresses to dial out to first

If you have the mode set to `ipv4_only` and there are no A records, the request will fail. Likewise for `ipv6_only`, if there's no AAAA records, the request will fail.

The preference is enforced at the Dial syscall (`tcp4`/`tcp6` or `udp4`/`udp6`), so the outbound connection never silently falls back to the other family. Unknown values are rejected at provision, and an `_only` preference paired with a `local_address` that has no matching-family entry is also rejected at provision (otherwise every source would be silently skipped).

## Use-case
I added this code to support a use-case of load-balancing outbound traffic to a host through multiple IPv4/IPv6 addresses. The current code always sends requests out through the primary IP of the server which isn't ideal for me. I assume this would be troublesome to others who need traffic to go out through the non-primary IP as well.

Secondly, during my initial implementation, I had noticed my requests were failing towards my upstream because the upstream only whitelisted the IPv4 address of my Caddy server, but Caddy was sending out requests over IPv6 which wasn't desired. So that's why I added `resolver_preference` as well.

## Backward compatibility
Configurations that set neither `local_address` nor `resolver_preference` are unaffected: the new code paths are gated on one of those being set, so the dial is identical to pre-PR behavior.

## Testing
I've tested for regressions on my local machine and validated the new functionality (incl. health checks) and everything seems to work as expected. I validated the resolver functionality using `ip4.me`, `ip6.me`, `ip6only.me`, exercising every valid combination of listener family × source family × resolver preference.

That being said, If I've missed something, please let me know.


docs/handlers/proxy.md
```diff
@@ -84,22 +83,34 @@ Load balacing options include `lb_policy`, `lb_try_duration` and `lb_try_interva
...
+
 - `dial` contains a list of network addresses to dial. Each address must be exactly 1 socket, e.g. `10.1.2.3:80`.
   No port ranges are currently supported by the handler. At least one dial address must be provided per upstream.
   Multiple addresses are dialled one by one until a connection is successfully established.  
 
+- `local_address` specifies the source address when connecting to the upstream(s). Provide only the address (no
+  protocol prefix); the protocol is inferred from each upstream dial target (TCP by default if none is specified).
+  If a port is included, that exact port is used; if omitted, the OS chooses an ephemeral port (:0). Unix
+  source paths are supported only for Unix upstreams; for TCP/UDP upstreams, a Unix `local_address` is invalid, and
+  for Unix upstreams an IPv4/v6 `local_address` is invalid. Multiple addresses may be provided as a comma-separated
```
vnxme: Caddyfile syntax provides for **space**-separated lists. Is there a reason for breaking the defaults?

docs/handlers/proxy.md
```diff
@@ -112,15 +123,16 @@ Each `upstream` has the following fields:
     | `tls_timeout`                              | `handshake_timeout`                                         |
     | `tls_trust_pool`                           | `ca`                                                        |
 
+Only two fields support [placeholders](https://caddyserver.com/docs/conventions#placeholders).
```
vnxme: Could we also support placeholders in `local_address`? I suppose the known ones can be expanded at provision, while the remaining ones must be resolved at handle -- same as for `dial`.

docs/handlers/proxy.md
```diff
@@ -143,6 +155,7 @@ proxy [<upstreams...>] {
     # multiple upstream options are supported
     upstream [<address:port>] {
         dial <address:port> [<address:port>]
+        local_addr <network/host[:port]>
```
vnxme: Should be `local_addr <address[:port]> [<address[:port]>]`, if we accept more than one address. The network part looks confusing, since the protocol is inferred from the dial target.

docs/handlers/proxy.md
```diff
@@ -143,6 +155,7 @@ proxy [<upstreams...>] {
...
         max_connections <int>
```
vnxme, line 180: `resolver_preference ipv4_first|ipv4_only|ipv6_first|ipv6_only` is missing here.

docs/handlers/proxy.md
```diff
@@ -84,22 +83,34 @@ Load balacing options include `lb_policy`, `lb_try_duration` and `lb_try_interva
...
 
 Each `upstream` has the following fields:
+
 - `dial` contains a list of network addresses to dial. Each address must be exactly 1 socket, e.g. `10.1.2.3:80`.
   No port ranges are currently supported by the handler. At least one dial address must be provided per upstream.
   Multiple addresses are dialled one by one until a connection is successfully established.  
 
+- `local_address` specifies the source address when connecting to the upstream(s). Provide only the address (no
+  protocol prefix); the protocol is inferred from each upstream dial target (TCP by default if none is specified).
+  If a port is included, that exact port is used; if omitted, the OS chooses an ephemeral port (:0). Unix
```
vnxme: Could you please explain how it should work for a Unix socket (`SOCK_STREAM`, `SOCK_SEQPACKET`, `SOCK_DGRAM`)? If `dial` contains `unix//tmp/example.sock`, what value can `local_addr` have?

modules/l4proxy/upstream.go
```diff
@@ -39,6 +39,15 @@ type Upstream struct {
...
 	Dial []string `json:"dial,omitempty"`
 
+	// Local address to bind to when making the request to upstream. Multiple addresses are supported as a comma-separated list.
+	// The first address matching the upstream’s address family (IPv4/IPv6/Unix) is used, otherwise the OS default is used.
+	// Source ports can be specified for both IPv6 and IPv6, but IPv6 must use brackets when specifying the port. i.e. [2001:db8::1]:12345.
+	LocalAddr string `json:"local_address,omitempty"`
+
+	// Preference for address family resolution. One of: "ipv4_only", "ipv6_only",
+	// "ipv4_first", "ipv6_first". Default is "ipv4_first".
+	ResolverPreference string `json:"resolver_preference,omitempty"`
```
vnxme, line 80: If this field supports a limited number of  pre-defined values, it's worth validating them on provision and/or unmarshalling. I mean a user should get an error if `ipv46_only` is entered by mistake, rather than simply falling back to `ipv4_first`.

modules/l4proxy/upstream.go
```diff
@@ -203,6 +214,21 @@ func (u *Upstream) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
 				return d.ArgErr()
 			}
 			shortcutArgs = append(shortcutArgs, d.RemainingArgs()...)
+		case "resolver_preference":
```
vnxme, line 325: `resolver_preference` should only occur once per upstream config, so I suggest we add the `hasResolverPreference` magic.

modules/l4proxy/upstream.go
```diff
@@ -203,6 +214,21 @@ func (u *Upstream) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
+			if d.CountRemainingArgs() != 1 {
+				return d.ArgErr()
+			}
+			d.NextArg()
+			u.ResolverPreference = d.Val()
+		case "local_addr":
+			if hasLocalAddr {
```
vnxme: Unlike the case above, why shall we limit the `local_addr` occurrence per upstream config? I.e. multiple local addresses can be entered in a single line (space-separated) or in a few lines, each starting with `local_addr` -- same as `dial`.

modules/l4proxy/upstream.go
```diff
@@ -170,6 +179,7 @@ func (u *Upstream) totalConns() int {
 //
 //	upstream [<address:port>] {
 //		dial <address:port> [<address:port>]
+//		local_addr <address:port>
```
vnxme: Same as above: `local_addr <address:port> [<address:port>]` and `resolver_reference ipv4_first|ipv4_only|ipv6_first|ipv6_only`.

vnxme (changes requested):
@JDTX0, thank you for contributing! The idea behind the changes you propose is well understood.

I've added a few high-level questions and comments to the code. It's also worth updating the relevant integration tests, by the way.

docs/handlers/proxy.md
JDTX0: Changed in https://github.com/mholt/caddy-l4/pull/361/changes/5f06933a4e84a296ec98cb25549a4e503a4c75ef

docs/handlers/proxy.md
JDTX0: Added support for placeholders in https://github.com/mholt/caddy-l4/pull/361/changes/5f06933a4e84a296ec98cb25549a4e503a4c75ef

docs/handlers/proxy.md
JDTX0: Yeah, you're right, I've adjusted this in https://github.com/mholt/caddy-l4/pull/361/changes/3d2769e75e6e77fb824b090f0f59ca8ee1fd71fe

docs/handlers/proxy.md
JDTX0: Added in https://github.com/mholt/caddy-l4/pull/361/changes/3d2769e75e6e77fb824b090f0f59ca8ee1fd71fe

docs/handlers/proxy.md
JDTX0: I only added support for it as I saw it was supported in Caddy itself when they added support for specifying local addresses: https://github.com/caddyserver/caddy/pull/6504/changes#diff-8839c76ab9a0e424d77cabc5357fbcdfb166d6b6cc7a076114fb24a5b16814b8R206

It provides no real value, so I've removed it in https://github.com/mholt/caddy-l4/pull/361/changes/1500d0ba4b9720092244c6fdb301d310f02f1f85

modules/l4proxy/upstream.go
JDTX0: Great point. I added strict validation for this in https://github.com/mholt/caddy-l4/pull/361/changes/77b28cdae0f041dc327e415459e1cf89ec936f9b

modules/l4proxy/upstream.go
JDTX0: Done, added in https://github.com/mholt/caddy-l4/pull/361/changes/77b28cdae0f041dc327e415459e1cf89ec936f9b

modules/l4proxy/upstream.go
JDTX0: I adjusted it to support this syntax in https://github.com/mholt/caddy-l4/pull/361/changes/77b28cdae0f041dc327e415459e1cf89ec936f9b

modules/l4proxy/upstream.go
JDTX0: Fixed in https://github.com/mholt/caddy-l4/pull/361/changes/77b28cdae0f041dc327e415459e1cf89ec936f9b#diff-3fed643c32c949ce1c4bace5373b17a74a4e92c49acd392e6b84fca1c4ed9197R218-R219

JDTX0:
@vnxme Thank you for the feedback and review! I've addressed your comments and fixed a number of other issues;

- Fixed an issue where setting `ipv4_only` or `ipv6_only` could fall back to another family. `net.Dial` was being called with `tcp` which in turn let Go's happy eyeballs choose a family that wasn't preferred. Now we narrow the Dial call by specifying tcp4/tcp6 based on the resolved address

- Outright reject problematic config combinations where users specify `ipv4_only` for a resolver preference but only provide an IPv6 `local_address`. Same for the other way around; If someone specifies ipv6_only for resolver pref, but gives an IPv4 `local_address`, it will also be rejected.

- Prevent `resolveDestFamily` from being called unconditionally. If neither `local_address` or `resolver_preference` is specified, then we don't perform the DNS lookup via `resolveDestFamily` so we don't introduce any possible performance regression for users who don't use either new option.

- Merged in latest code from master, resolved the conflict and ensured my code handles the new nullable `NetworkAddress` that was introduced when support for dynamic port numbers was added

I've also added the integration tests and updated the proxy tests. I compiled Caddy and tested all of the possible scenarios using `ip6.me`, `ip4.me`, and `ip6only.me` and ensured everything works as I expect with the resolver preferences and local addresses being set to all possible combinations.

Should be ready for another round of review.

docs/handlers/proxy.md
```diff
@@ -262,9 +297,121 @@ filled at random:
 }
 ```
 
+#### Controlling the outbound source address and resolver family
```
vnxme: I suggest we move this section to the end of this file, so that the JSON section could immediately follow the Caddyfile section.

docs/handlers/proxy.md
```diff
@@ -262,9 +297,121 @@ filled at random:
...
+
+`local_address` pins the source address Caddy uses when dialing an upstream, and `resolver_preference`
+restricts or biases which DNS family is used when the upstream is a hostname. Typical use-cases:
+
+- **Dual-stack outbound load-balancing.** Provide an IPv4 and an IPv6 source; Caddy picks the one that
+  matches the peer it is about to dial. This lets you spread outbound traffic across multiple
+  source IPs per family instead of always egressing from the host's primary IP.
+- **IP-whitelisted upstreams.** If a remote endpoint only allows a specific source IP (or only one
```
vnxme: Above you keep 1 empty line between list items. Shall we add an empty line here as well?

vnxme (changes requested):
@JDTX0, thank you for implementing the changes discussed above. Overall, I have no objections to the code, just a few minor comments to take into account.

@mholt, could you please take a look if there is anything else you would like to improve in this PR before we may merge it?

mholt (approved):
Yeah I think this is fine! Niche but fine :)

docs/handlers/proxy.md
JDTX0: Fixed in https://github.com/mholt/caddy-l4/pull/361/commits/4994eb0f065de26baa6afde1c34b04898864aadc

docs/handlers/proxy.md
JDTX0: Fixed in https://github.com/mholt/caddy-l4/pull/361/commits/4994eb0f065de26baa6afde1c34b04898864aadc

docs/handlers/proxy.md
```diff
@@ -262,9 +299,109 @@ filled at random:
...
+                            ]
+                        }
+                    ]
+                }
+            }
+        }
+    }
+}
+```
+
```
vnxme: You've moved the description text to the end of the file, but left these configuration examples here - why? I mean, there is a general-purpose Caddyfile example starting at line 259, which should be followed by its JSON equivalent starting at line 405. Your examples starting at lines 302 and 340 are specific to "controlling the outbound ..." block, so I suggest we move them to the end as well.

vnxme (changes requested):

docs/handlers/proxy.md
JDTX0: d'oh. resolved in https://github.com/mholt/caddy-l4/pull/361/commits/ff306b64834d53563b21086a8c5803e690db29f1

vnxme (approved):
@JDTX0, thank you for working on it! Looks like we're ready to merge.

Issue #371, opened 2026-01-19
Closing a connection on a layer4 route


Niddrym:
I've been reading through the examples and the documentation, and I haven't found a way on l4 to abort/close a connection (though it's quite possible I've missed it).  Is it possible to abort a connection on a specific route in l4? 

I've included an example config below.

```
{
    servers {
        listener_wrappers {
            layer4 {
                #ssh over tls
                @tls-ssh tls sni ssh.example.com
                route @tls-ssh {
                    tls {
                        connection_policy {
                            alpn http/1.1
                        }
                    }
                    subroute {
                        @ssh-tls ssh
                        route @ssh-tls {
                            proxy 127.0.0.1:22
                        }
                        route  {
                            # Close connection here
                        }
                    }
                }
                #catch raw ssh connections
                @ssh ssh
                route @ssh {
                    proxy 127.0.0.1:22
                }
            }
            #fallback to regular tls sites
            tls
        }
    }
}
```

PR #373, opened 2026-01-27
Implement close handler

vnxme:
This PR implements `close` handler as requested in #371. As mentioned in the docs, the handler is relevant to connection-oriented protocols (e.g. TCP) and irrelevant to connectionless protocols (e.g. UDP).

I haven't tested this handler yet, so please report whether it works correctly for you with TCP and doesn't break UDP.


mholt:
Hmm yeah that does seem like a good idea Francis, good thinking.

PR #381, opened 2026-02-12
Listener wrapper clarification

### LINKED ISSUE #375: Using both TLS SNI and Caddy services on SSL causes intermittent issues
### 1a. Operating system and version

Debian 13 with Docker

### 1b. Caddy version (run `caddy version` or paste commit SHA)

```
v2.10.2 h1:g/gTYjGMD0dec+UgMw8SnfmJ3I9+M2TdvoRL/Ovu6U8=
```

```
Caddy 2.10.2
caddy-docker-proxy: 41252135f9b14801713ddb3f6a458b43c7d69c2c
caddy-l4: 040d25cc886ab41afe5a3e25a7cb33a2fcafa202
```

### 2a. What happens (briefly explain what is wrong)

Apologies for this if I just misunderstood something or got something wrong.

I'm using Caddy to reverse proxy multiple services on a server, including both LANCache and Docker registry. In order to handle https for the LANCache domains, I'm using layer4 with TLS SNI to route them to their real host. You can see my Caddy config in the example section below.

However, with the layer4 section added, I begin to get intermittent issues from the Docker registry and other SSL services behind Caddy. Docker fails to push images with an EOF error. Curl also fails to get pages about 50% of the time, I can spam the command and get different results. These failures happen both for reverse proxied services and the domains getting routed through layer4 TLS SNI.

### 2b. Why it's a bug (if it's not obvious)

The intermittent, sometimes working sometimes not behavior makes me think that I misunderstood something, or this is a bug. I can visit https://docker.mydomain.com/v2 or cdn2.unrealengine.com in Firefox and get a 200 response, whereas Docker CLI always fails and Curl is inconsistent, sometimes succeeding, sometimes not.

### 2c. Log output

Caddy did not output any logs about connections related to this issue. Docker Registry would sometimes log on the server side after many attempts that there was an EOF error.

## 3. Tutorial (minimal steps to reproduce the bug)

Trimmed down example of my Caddy config
```
layer4 {
  :443 {
    @lancache_domains tls sni *.unrealengine.com
    route @lancache_domains {
      proxy tcp/{l4.tls.server_name}:443
    }
  }
}
docker.mydomain.com {
  reverse_proxy registry:5000
}
```

Docker images can no longer be pushed with the layer4 section
```
sudo docker push docker.mydomain.com/hello-world
Using default tag: latest
The push refers to repository [docker.mydomain.com/hello-world]
Get "https://docker.mydomain.com/v2/": EOF
```

Curl is inconsistent, sometimes succeeding, sometimes failing
```
curl -v -k https://mydomain.com
* Host mydomain.com:443 was resolved.
* IPv6: (none)
* IPv4: xxx.xxx.xxx.xxx
*   Trying xxx.xxx.xxx.xxx:443...
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* SSL Trust Anchors:
*   CAfile: /etc/ssl/certs/ca-certificates.crt
* TLSv1.3 (OUT), TLS alert, decode error (562):
* TLS connect error: error:0A000126:SSL routines::unexpected eof while reading
* closing connection #0
curl: (35) TLS connect error: error:0A000126:SSL routines::unexpected eof while reading
```

MarcelWaldvogel:
Here is an example to kind of fix #375 

(Also fixes another slight documentation issue I ran across)

### DIFF
--- a/docs/handlers/proxy.md
+++ b/docs/handlers/proxy.md
@@ -262,6 +262,36 @@ filled at random:
 }
 ```
 
+An example config of the Layer 4 app that allows mixing of TLS termination and
+transparent TLS forwarding. `proxy_protocol` and `http_redirect` are optional:
+```caddyfile
+{
+    servers :443 {
+        listener_wrappers {
+            layer4 {
+                @transparent tls sni transparent.example.com
+                route @transparent {
+                    proxy transparent:443 {
+                        proxy_protocol v2
+                    }
+                }
+                http_redirect
+                tls
+            }
+        }
+    }
+}
+
+# For redirects and working ACME
+http://transparent.example.com {
+    reverse_proxy http://transparent
+}
+terminate.example.com {
+    reverse_proxy http://backend
+}
+```
+
+
 ### JSON
 
 JSON equivalent to the caddyfile config provided above:


mholt (approved):
Thanks!

docs/handlers/proxy.md
```diff
@@ -262,6 +262,36 @@ filled at random:
 }
 ```
 
+An example config of the Layer 4 app that allows mixing of TLS termination and
```
vnxme: I don't think this example is strictly related to the `proxy` handler, so I would rather move it out from `docs/handlers/proxy.md` to `docs/examples/[short-name].md`. We may have a link to this example config somewhere, but I don't know whether it should be the `proxy` handler page either.

docs/handlers/proxy.md
```diff
@@ -262,6 +262,36 @@ filled at random:
...
+                }
+                http_redirect
+                tls
+            }
+        }
+    }
+}
+
+# For redirects and working ACME
+http://transparent.example.com {
```
vnxme: @mholt, @francislavoie, need your help here. Does Caddy obtain a TLS certificate for the domain referenced in an HTTP (not HTTPS) block?

I assume we should have `https://transparent.example.com` here to ensure Caddy has the relevant TLS certificate, or adjust line 289 to include `*.example.com` to use a wildcard certificate instead.

vnxme (commented):
Thanks for the corrections to `docs/servers.md`. I added a few questions/suggestions to `docs/handlers/proxy.md`.

I also mentioned [here](https://github.com/mholt/caddy-l4/issues/375#issuecomment-3895117170) the note relevant to the issue described in #375 that already exists in the docs -- we might probably make it more visible.

By the way, we already have a [listener wrapper example](https://github.com/mholt/caddy-l4/blob/master/docs/examples/postgres-over-tls.md) in the docs, though it's just more specific (related to PostgreSQL) than a general HTTPS proxy.

PR #403, opened 2026-04-02
lint: Modernize atomic operations in the proxy handler

vnxme:
This PR fixes the following linter warnings:
```
modules\l4proxy\upstream.go:367:2: atomic: var numConns int32 may be simplified using atomic.Int32 (modernize)
        numConns  int32
        ^
modules\l4proxy\upstream.go:368:2: atomic: var unhealthy int32 may be simplified using atomic.Int32 (modernize)
        unhealthy int32
        ^
modules\l4proxy\upstream.go:369:2: atomic: var fails int32 may be simplified using atomic.Int32 (modernize)
        fails     int32
        ^
3 issues:
* modernize: 3
```

No AI has been involved.

### DIFF
--- a/modules/l4proxy/upstream.go
+++ b/modules/l4proxy/upstream.go
@@ -134,7 +134,7 @@ func (u *Upstream) healthy() bool {
 	}
 	if u.healthCheckPolicy != nil && u.healthCheckPolicy.MaxFails > 0 {
 		for _, p := range u.peers {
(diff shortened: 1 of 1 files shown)


vnxme:
I assume this PR is safe to merge because it only updates the existing atomic operations to follow the approach introduced in Go 1.19.

PR #414, opened 2026-05-02
l4quic: panic-recover the WaitForMore goroutine

kkroo:
## Summary

If \`WaitForMore\` (or any code reachable from it via a future refactor that breaks the \`cx.isPacketConn\` invariant) panics, the goroutine's \`close(done)\` never runs and \`Match()\` blocks forever at the \`<-done\` wait below it.

## Fix

Wrap \`close(done)\` in a deferred panic recovery, and log the panic with the remote addr so operators can attribute the symptom.

Defensive only — no known panic source today; this is a robustness fix to prevent a future change in \`WaitForMore\` (or anything it calls into) from silently deadlocking the matcher in production.

## Test plan

- [x] \`go test ./modules/l4quic/\` passes (existing tests cover the success and timeout-by-deadline paths; the panic-recovery path can't be exercised without a synthetic panic injection).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/l4quic/matcher.go
+++ b/modules/l4quic/matcher.go
@@ -35,6 +35,7 @@ import (
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
 	"github.com/caddyserver/caddy/v2/modules/caddytls"
 	"github.com/quic-go/quic-go"
+	"go.uber.org/zap"
 
 	"github.com/mholt/caddy-l4/layer4"
 	"github.com/mholt/caddy-l4/modules/l4tls"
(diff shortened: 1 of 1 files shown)


modules/l4quic/matcher.go
```diff
@@ -170,14 +171,25 @@ func (m *MatchQUIC) Match(cx *layer4.Connection) (bool, error) {
...
+		// the <-done wait below. Logs the panic + remote addr so
+		// operators can attribute the symptom.
+		defer func() {
+			if r := recover(); r != nil {
+				cx.Logger.Error("panic in quic matcher WaitForMore goroutine",
+					zap.Any("panic", r),
+					zap.String("remote", cx.RemoteAddr().String()))
+			}
+			close(done)
+		}()
```
vnxme, line 199:
```suggestion
		// close(done) lives in a deferred recovery so a future refactor
		// that breaks the cx.isPacketConn invariant (or any other panic
		// inside WaitForMore) cannot leave Match() blocked forever at
		// the <-done wait below. Logs the panic + remote addr so
		// operators can attribute the symptom.
		defer func() {
			if r := recover(); r != nil {
				cx.Logger.Error("panic in quic matcher WaitForMore goroutine",
					zap.String("network", cx.LocalAddr().Network()),
					zap.String("local", cx.LocalAddr().String()),
					zap.String("remote", cx.RemoteAddr().String()),
					zap.Any("panic", r),
				)
			}
			close(done)
		}()
```

vnxme (changes requested):
Hi, thanks for proposing these code changes! I don't know if we might expect any panic inside WaitForMore (please share if you have any reproducible config), but it looks OK. I suggest we reorder log fields and add network and local address as well.

kkroo:
Thanks @vnxme — applied the suggested field reordering plus network+local addresses. Force-pushed; please re-review.

For the panic-source question on #414: no specific reproducer in hand — it's defensive, motivated by the long-tail risk of a future change inside `WaitForMore` (or its callees) breaking the assumption that `close(done)` always runs. Happy to drop the PR if you'd rather wait for a real failure mode to emerge.

PR #415, opened 2026-05-02
l4quic: log Accept() errors at WARN/DEBUG split

kkroo:
## Summary

The \`Accept()\` error handler in \`MatchQUIC.Match()\` returns \`ErrConsumedAllPrefetchedBytes\` for every error class with no log trail. In production it's useful to distinguish:

- **Expected:** \`context.DeadlineExceeded\` (\`QUICAcceptTimeout\` fired) and \`context.Canceled\` (\`WaitForMore\` goroutine cancelled \`qContext\`). These happen on every non-handshaking UDP packet, so they're DEBUG.
- **Unexpected:** transport closed, listener closed, future quic-go variants. These usually indicate a real transport bug — log at WARN with the remote addr.

Purely additive; no change to the return value or matcher semantics.

## Test plan

- [x] \`go test ./modules/l4quic/\` passes (no behavior change to assert against; logs are diagnostic).

## Relation to mholt/caddy-l4#413

Independent. #413 changes the deadline-path return value to \`(false, nil)\` to enable fall-through. This PR is purely additive logging that's useful regardless of which return-value semantics are in effect.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/l4quic/matcher.go
+++ b/modules/l4quic/matcher.go
@@ -35,6 +35,7 @@ import (
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
 	"github.com/caddyserver/caddy/v2/modules/caddytls"
 	"github.com/quic-go/quic-go"
+	"go.uber.org/zap"
 
 	"github.com/mholt/caddy-l4/layer4"
 	"github.com/mholt/caddy-l4/modules/l4tls"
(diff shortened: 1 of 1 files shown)


modules/l4quic/matcher.go
```diff
@@ -186,7 +187,23 @@ func (m *MatchQUIC) Match(cx *layer4.Connection) (bool, error) {
...
+		// timeout/cancel race.
+		if !errors.Is(err, context.DeadlineExceeded) && !errors.Is(err, context.Canceled) {
+			cx.Logger.Warn("quic matcher: unexpected Accept() error",
+				zap.String("remote", cx.RemoteAddr().String()),
+				zap.Error(err))
+		} else {
+			cx.Logger.Debug("quic matcher: Accept() timeout or cancel",
+				zap.String("remote", cx.RemoteAddr().String()),
+				zap.Error(err))
+		}
```
vnxme, line 215:
```suggestion
		if !errors.Is(err, context.DeadlineExceeded) && !errors.Is(err, context.Canceled) {
			cx.Logger.Warn("quic matcher: unexpected Accept() error",
				zap.String("network", cx.LocalAddr().Network()),
				zap.String("local", cx.LocalAddr().String()),
				zap.String("remote", cx.RemoteAddr().String()),
				zap.Error(err))
		} else {
			cx.Logger.Debug("quic matcher: Accept() timeout or cancel",
				zap.String("network", cx.LocalAddr().Network()),
				zap.String("local", cx.LocalAddr().String()),
				zap.String("remote", cx.RemoteAddr().String()),
				zap.Error(err))
		}
```

vnxme (changes requested):
I suggest we log additional fields (network, local).

kkroo:
Thanks @vnxme — applied the suggested network+local fields. Force-pushed; please re-review.

For the panic-source question on #414: no specific reproducer in hand — it's defensive, motivated by the long-tail risk of a future change inside `WaitForMore` (or its callees) breaking the assumption that `close(done)` always runs. Happy to drop the PR if you'd rather wait for a real failure mode to emerge.

PR #423, opened 2026-06-01
l4proxy: add active HTTP health checks

### LINKED ISSUE #203: http active health check
add a way to check upstreams via http request 

defined a endpoint on my host /healthy which returns ok on success currently i used haproxy but i want switch to caddy

example config in http health check:

```json
"active": {
			"port": 0,
			"interval": 0,
			"timeout": 0,
			
			// new fields
			"type": "http",
			"host": "my-host.com",
                        "path": "/health",
                        "expect_status": 200,
			
		},
```

tannevaled:
## What

Adds an optional HTTP mode to the layer4 proxy's active health checks. Today the active health check only performs a raw TCP dial, so it can tell whether a port is open but nothing about the application-level state behind it.

New `ActiveHealthChecks` fields (JSON):
- `uri` — request path; when set, the check performs an HTTP `GET` instead of a TCP dial
- `expect_status` — healthy status code, or a class (e.g. `2` matches any 2xx); default `200`
- `https` — perform the check over TLS
- `tls_skip_verify` — skip TLS verification (useful with self-signed certs)

Caddyfile equivalents: `health_uri`, `health_status` (accepts `200` or `2xx`), `health_https`, `health_tls_skip_verify`.

When `uri` is unset the check stays a raw TCP dial, so existing behavior is unchanged.

## Why

It lets a layer4 proxy follow an application-level signal instead of mere port-liveness. For example, Patroni (PostgreSQL HA) exposes `/primary`, which returns `200` only on the elected leader; with this change the proxy can route to whichever upstream is currently primary and follow failover automatically — something the TCP-only check cannot do, since every node accepts TCP connections regardless of role.

## Tests

Added `healthchecks_test.go`: healthy/unhealthy by status, status-class match, HTTPS (self-signed), connection-refused, invalid-URI, and Caddyfile parsing (happy + error paths). `go test ./modules/l4proxy/` passes; `gofmt` and `go vet` are clean.

### DIFF
--- a/docs/handlers/proxy.md
+++ b/docs/handlers/proxy.md
@@ -56,6 +56,22 @@ Active health check options include `health_interval`, `health_port` and `health
 - `health_rise` is the number of consecutive successful active health checks required to mark an unhealthy upstream
   healthy again (by default, `1`).
 
+By default an active health check is a raw TCP dial. It can instead probe over HTTP, which lets the proxy follow an
+application-level signal (e.g. a primary-election endpoint such as Patroni's `/primary`) rather than mere port
+liveness. The HTTP probe is enabled by setting `health_uri` and is configured with these options, which correspond to
+the `uri`, `expect_status`, `https` and `tls_skip_verify` fields of the `l4proxy.ActiveHealthChecks` structure:
+
+- `health_uri` is the request path for an HTTP `GET` health check (e.g. `/primary`). When set, the check is performed
+  over HTTP (on `health_port` if configured, otherwise the dial port) instead of a raw TCP dial;
+
+- `health_status` is the status code considered healthy (by default, `200`). A value below `100` is treated as a
+  status class, e.g. `2` matches any `2xx` response;
+
+- `health_https` performs the HTTP health check over TLS;
+
+- `health_tls_skip_verify` disables TLS certificate verification for an `health_https` check (useful with self-signed
+  certificates).
+
 **Passive health checks** monitor proxied connections for errors or timeouts. To minimally enable passive health checks,
 set `passive` field equal to an empty structure inside `health_checks` in a JSON configuration or include any passive
 health check option into a Caddyfile.
--- a/integration/caddyfile_adapt/gd_handler_proxy_health_http.caddytest
+++ b/integration/caddyfile_adapt/gd_handler_proxy_health_http.caddytest
@@ -1,0 +1,54 @@
+{
+	layer4 {
+		:5432 {
+			route {
+				proxy upstream.local:5432 {
+					health_port 8008
+					health_uri /primary
+					health_status 2xx
+					health_https
+					health_tls_skip_verify
+				}
+			}
+		}
+	}
+}
+----------
+{
+	"apps": {
+		"layer4": {
+			"servers": {
+				"srv0": {
+					"listen": [
+						":5432"
+					],
+					"routes": [
+						{
+							"handle": [
+								{
+									"handler": "proxy",
+									"health_checks": {
+										"active": {
+											"expect_status": 2,
+											"https": true,
+											"port": 8008,
+											"tls_skip_verify": true,
+											"uri": "/primary"
+										}
+									},
+									"upstreams": [
+										{
+											"dial": [
+												"upstream.local:5432"
+											]
+										}
+									]
+								}
+							]
+						}
+					]
+				}
+			}
+		}
+	}
+}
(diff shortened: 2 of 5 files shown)


tannevaled:
@vnxme no rush at all — just a gentle flag that this PR and the stacked #426 are green and rebased onto the #446 panic fix, so they're ready whenever you have the bandwidth. Same for the dynamic-upstreams pair (#429 → #430) if it's convenient to look at them together. Thanks again for working through the series so quickly!

vnxme:
@tannevaled, thank you for reminder. These 4 remaining PRs in your `proxy`-related series are far more complex.

Before diving into the code, let's start with the following questions:
- The functionality this PR deals with is directly related to what we have in [caddyhttp/reverseproxy/healthcheck.go](https://github.com/caddyserver/caddy/blob/945d199724f17f313c7fbc90eceba70fe37f47e6/modules/caddyhttp/reverseproxy/healthchecks.go#L70-L134) and what is covered in the docs [here](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#active-health-checks). Is it possible we reuse the mainline code? If necessary, we could probably make some mainline functions/structures exportable, so that they could be reused by the layer4 app.
- This PR introduces `URI` and `ExpectStatus` fields which directly match the mainline code, at least as far as their names are concerned. Is my understanding correct that you consider these two fields crucial, that's why you didn't include all the remaining fields that Caddy supports?
- This PR also includes `HTTPS` and `TLSSkipVerify` fields that are missing in Caddy. As for the former one, I don't understand its role at all if `URI` contains `https://example.com` or `http://website.local`. What's your logic behind this field? Could it safely be omitted in favour of parsing the `URI` field?
- `TLSSkipVerify` has two issues for me: (a) its name is confusing because it doesn't match the `tls_insecure_skip_verify` option of the `proxy` handler; (b) it seems to have no counterpart in the mainline code. Issue (a) is pretty easy -- we could just rename the field. When it comes to issue (b), I would prefer we extend the mainline code, i.e. add this option to the `reverseproxy` module and reuse it here afterwards. Would it work for you?

Please note that the approach you originally took in this PR may also be OK, but I'm trying to review the changes from the post-merger integration standpoint. Someday we will merge this app into Caddy, and the fewer places in the code we'll have to rewrite the better.

tannevaled:
Thanks — and I'm fully on board with reviewing this from the "someday this merges into Caddy" angle; minimizing divergence from mainline is the right lens. I dug into `caddyhttp/reverseproxy/healthchecks.go` to answer concretely:

**1. Reuse mainline code.** The doer (`(*Handler).doActiveHealthCheck`) is coupled to reverseproxy's `*Handler` / `*Upstream` / `DialInfo` / `Transport`, so it can't be imported as-is. But the transport-neutral part — build the request from URI/method/headers/body, GET it, then apply `ExpectStatus`/`ExpectBody`/`MaxSize` — is cleanly extractable. I'd suggest the same move we made for DNS discovery in caddyserver/caddy#7790 (which landed under `modules/internal/network`): factor the HTTP probe into an internal helper in core that both `reverseproxy` and a future in-tree layer4 can call, and have this PR call the same logic. Happy to open that core PR if you like the direction.

**2. `URI` / `ExpectStatus` subset.** Correct — I picked the minimum needed for HAProxy-style HTTP checks. Glad to align the field set and names/JSON tags 1:1 with mainline (`uri`, `expect_status`, `expect_body`, `headers`, plus `method`, `body`, `follow_redirects`, `max_size`) so it's a drop-in.

**3. `HTTPS` field.** You're right, it's redundant. Mainline carries no `https` bool — it defaults the scheme to `http` and lets the transport override it (`HTTPTransport.OverrideHealthCheckScheme` → `https` when the upstream transport has TLS). The layer4 proxy already knows this from `Upstream.TLS`, so I'll drop `HTTPS` and derive the scheme from whether the upstream is TLS — same behaviour as mainline (the check speaks the same scheme as the real traffic).

**4. `TLSSkipVerify`.** Agreed on both counts, and I think we can avoid touching mainline entirely: mainline gets skip-verify from the transport's TLS config, and `Upstream.TLS` here is a `reverseproxy.TLSConfig`, which already has `insecure_skip_verify`. So rather than rename the field, I'll drop it and use `Upstream.TLS.InsecureSkipVerify` — the probe then trusts the upstream cert exactly like the proxied connection does. No new option, no divergence.

Net effect: `URI`/`ExpectStatus`/`ExpectBody`/`Headers` stay and match mainline names; `HTTPS` and `TLSSkipVerify` go away, derived from the existing `Upstream.TLS`. Does that direction work for you? If so I'll rework #423 along these lines (and fold #426 in).

PR #424, opened 2026-06-01
l4proxy: add Caddyfile unmarshaling tests

tannevaled:
## What

Adds unit tests for the proxy handler's `UnmarshalCaddyfile`, which previously had no coverage:

- a full configuration exercising active + passive health checks, `lb_policy` / `lb_try_duration` / `lb_try_interval`, `proxy_protocol`, and both inline and block `upstream` forms;
- the common error paths: duplicate options, invalid numeric/duration values, an unknown load-balancing policy, and an unknown directive.

## Why

`UnmarshalCaddyfile` was at 0% coverage. These tests bring it to ~72% and guard the Caddyfile surface against regressions. No production code changes.

## Tests

`go test ./modules/l4proxy/` passes; `gofmt` / `go vet` / `golangci-lint` clean.

### DIFF
--- a/modules/l4proxy/caddyfile_test.go
+++ b/modules/l4proxy/caddyfile_test.go
@@ -16,79 +16,15 @@ package l4proxy
 
 import (
 	"testing"
-	"time"
 
-	"github.com/caddyserver/caddy/v2"
 	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
 )
 
-func TestUnmarshalCaddyfileFull(t *testing.T) {
-	d := caddyfile.NewTestDispenser(`proxy localhost:5432 udp/localhost:5433 {
-		health_interval 5s
-		health_port 8008
-		health_timeout 2s
-		fail_duration 30s
-		max_fails 3
-		unhealthy_connection_count 10
-		lb_policy round_robin
-		lb_try_duration 10s
-		lb_try_interval 1s
-		proxy_protocol v2
-		upstream localhost:5434
-	}`)
-
-	h := new(Handler)
-	if err := h.UnmarshalCaddyfile(d); err != nil {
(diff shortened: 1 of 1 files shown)


tannevaled:
Gentle nudge on this one whenever you have a spare moment — it's test-only (coverage for the existing Caddyfile unmarshaling, no behavior or API changes), so it should be a low-risk merge and independent of the feature PRs in the series. No urgency at all; happy to rebase if it's drifted.

vnxme:
@tannevaled, could you please explain in what way this test is better than [this file](https://github.com/mholt/caddy-l4/blob/master/integration/caddyfile_adapt/gd_handler_proxy.caddytest) and other integration tests we already have? Do we really need a separate test file here when we have a generic testing mechanism for Caddyfile unmarshalling? I think your coverage numbers may be inaccurate because they ignore these integration tests.

tannevaled:
You're right on both counts, thank you — and I've just pushed a change accordingly.

The `TestUnmarshalCaddyfileFull` happy-path test did duplicate what `gd_handler_proxy.caddytest` and the other adapt tests already verify. And your point about the coverage numbers is correct: I measured with `go test ./modules/l4proxy/`, which doesn't credit the `integration` package's adapt tests (they live in a separate package, so counting them needs `-coverpkg`) — so the redundancy wasn't visible in my numbers. My mistake.

I dropped that test. What remains is only `TestUnmarshalCaddyfileErrors`: the malformed/duplicate/unknown-directive cases that make `UnmarshalCaddyfile` return an error. Those error branches can't be reached by the adapt harness, since it only asserts that a valid Caddyfile adapts to expected JSON — there's no "expect adapt to fail" variant, and no `.caddytest` expresses a failure.

If you'd still rather not carry these as a separate Go test, I'm perfectly happy to close the PR — no attachment to it.

PR #425, opened 2026-06-01
l4proxy: optionally close connections when a peer goes unhealthy

tannevaled:
## What

Adds an opt-in `HealthChecks.CloseConnectionsOnUnhealthy` (Caddyfile: `close_connections_on_unhealthy`). When enabled, a peer's currently-open proxied connections are force-closed as soon as an **active** health check marks it unhealthy, instead of being left to run until they close on their own.

Closing the upstream side of each tracked connection makes the existing `io.Copy` loops in `proxy()` return, so the session tears down and the client reconnects — and is then routed to a healthy peer.

- Default is `false`, so existing behavior is unchanged.
- Per-peer connections are tracked (a small `map[net.Conn]struct{}` under a mutex) **only** when the option is enabled, so there's no overhead otherwise.
- `upConns[i]` corresponds to `upstream.peers[i]` (dialPeers dials one connection per peer, in order), which is how connections are associated with peers for tracking/untracking.

## Why

Pairs naturally with active health checks for failover scenarios. For example, with an HTTP health check against a PostgreSQL/Patroni `/primary` endpoint, when the primary is demoted its health check flips to unhealthy; this option then drops the stale sessions immediately rather than leaving clients pinned to a now read-only/blocked backend.

## Scope / open questions

- Triggered on the **active** health-check unhealthy transition. I deliberately left the passive path out of this PR to keep it focused; happy to extend if you'd prefer it also fire there.
- Open to naming/placement feedback (field on `HealthChecks` vs `PassiveHealthChecks`/`ActiveHealthChecks`).

## Tests

`closeconns_test.go`: peer track/untrack/close mechanics, active-check closes conns when enabled, leaves them when disabled, and Caddyfile parsing (happy + arg error). `go test ./modules/l4proxy/` passes; `gofmt` / `go vet` / `golangci-lint` clean.

### DIFF
--- a/docs/handlers/proxy.md
+++ b/docs/handlers/proxy.md
@@ -72,6 +72,12 @@ the similarly named fields of the `l4proxy.PassiveHealthChecks` structure:
 - `unhealthy_connection_count` limits the number of simultaneous connections to this upstream by marking it as "down"
   if it has this many or more concurrent connections.
 
+In addition, a `health_checks`-level option `close_connections_on_unhealthy` (corresponding to the
+`close_connections_on_unhealthy` field of `l4proxy.HealthChecks`) force-closes a peer's currently-open proxied
+connections the moment an active health check marks it unhealthy, instead of letting them run until they close on
+their own. This is useful for failover (e.g. moving clients off a demoted database primary). By default it is off,
+preserving the existing behavior.
+
 **Load balancing** distributes connections between upstreams. To minimally enable load balancing, set `load_balancing`
 field equal to an empty structure in a JSON configuration or include any load balancing option into a Caddyfile. Note:
 load balancing makes sense only if the handler has two or more upstreams.
--- a/docs/handlers/proxy.md
+++ b/docs/handlers/proxy.md
@@ -72,10 +72,11 @@ the similarly named fields of the `l4proxy.PassiveHealthChecks` structure:
 - `unhealthy_connection_count` limits the number of simultaneous connections to this upstream by marking it as "down"
   if it has this many or more concurrent connections.
 
-In addition, a `health_checks`-level option `close_connections_on_unhealthy` (corresponding to the
-`close_connections_on_unhealthy` field of `l4proxy.HealthChecks`) force-closes a peer's currently-open proxied
-connections the moment an active health check marks it unhealthy, instead of letting them run until they close on
-their own. This is useful for failover (e.g. moving clients off a demoted database primary). By default it is off,
+In addition, the active health check option `close_if_unhealthy` (corresponding to the `close_if_unhealthy` field of
+`l4proxy.ActiveHealthChecks`) force-closes a peer's currently-open proxied connections the moment an active health
+check marks it unhealthy, instead of letting them run until they close on their own. This is useful for failover
+(e.g. moving clients off a demoted database primary). It applies to active health checks only, which provide a clear
+healthy→unhealthy transition to act on; passive health checking has no equivalent event. By default it is off,
 preserving the existing behavior.
 
 **Load balancing** distributes connections between upstreams. To minimally enable load balancing, set `load_balancing`
--- a/modules/l4proxy/healthchecks.go
+++ b/modules/l4proxy/healthchecks.go
@@ -37,14 +37,6 @@ type HealthChecks struct {
 	// To minimally enable passive health checks, specify at least an empty
 	// config object.
 	Passive *PassiveHealthChecks `json:"passive,omitempty"`
-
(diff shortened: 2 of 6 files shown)


modules/l4proxy/proxy.go
```diff
@@ -680,6 +696,14 @@ func (h *Handler) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
...
 			}
 			h.HealthChecks.Passive.UnhealthyConnectionCount, hasUnhealthyConnCount = int(val), true
+		case "close_connections_on_unhealthy":
+			if d.CountRemainingArgs() != 0 {
+				return d.ArgErr()
+			}
+			if h.HealthChecks == nil {
+				h.HealthChecks = &HealthChecks{}
+			}
+			h.HealthChecks.CloseConnectionsOnUnhealthy = true
```
vnxme: This code allows setting this option twice/multiple times. Please add a guard to avoid duplicate options -- see an example a few lines above.

modules/l4proxy/healthchecks.go
```diff
@@ -37,6 +37,14 @@ type HealthChecks struct {
...
 	// config object.
 	Passive *PassiveHealthChecks `json:"passive,omitempty"`
+
+	// CloseConnectionsOnUnhealthy, when true, force-closes a peer's currently
+	// open proxied connections the moment an active health check marks it
+	// unhealthy, instead of letting them run until they close on their own.
+	// This is useful for failover, where clients should be moved off a backend
+	// as soon as it goes down (e.g. a demoted database primary). Default false,
+	// preserving the existing behavior.
+	CloseConnectionsOnUnhealthy bool `json:"close_connections_on_unhealthy,omitempty"`
```
vnxme: If this option is only relevant to active health checks, why do you suggest putting it inside the general health checks block that would imply it works for passive ones as well?

On the other hand, what's the logic for leaving the passive path out of this PR? Why do you believe this option should close connections to unhealthy peers only if they are actively monitored?

docs/handlers/proxy.md
```diff
@@ -72,6 +72,12 @@ the similarly named fields of the `l4proxy.PassiveHealthChecks` structure:
 - `unhealthy_connection_count` limits the number of simultaneous connections to this upstream by marking it as "down"
   if it has this many or more concurrent connections.
 
+In addition, a `health_checks`-level option `close_connections_on_unhealthy` (corresponding to the
+`close_connections_on_unhealthy` field of `l4proxy.HealthChecks`) force-closes a peer's currently-open proxied
+connections the moment an active health check marks it unhealthy, instead of letting them run until they close on
+their own. This is useful for failover (e.g. moving clients off a demoted database primary). By default it is off,
+preserving the existing behavior.
+
```
vnxme: What if we call this option `close_if_unhealthy`? I know, this is a matter of preference, and I would prefer to have as concise option names as possible -- in my view, fewer symbols reduce probability of typos and improve readability of config files. But I'm certainly open to discussion.

vnxme (commented):
Thank you for submitting this PR. Please see a few questions and comments to the code.

tannevaled:
Thanks for the review — all three addressed in the latest push.

1. **Duplicate guard** — added a `hasCloseIfUnhealthy` guard so specifying it twice now returns an error, matching the other options. Also added a duplicate-directive test case.

2. **Placement / why active-only** — you're right that putting it in the general `HealthChecks` block was misleading. I moved it into `ActiveHealthChecks`, where it actually applies. On leaving passive out: active checks give a clear healthy→unhealthy *transition* (`setHealthy(false)` returning `swapped`) that's a natural hook for closing connections. Passive health checking has no equivalent event — a peer isn't flipped by a transition, it's just considered "down" on demand once its failure count crosses the threshold during normal proxying. There's no single moment to trigger a close, so I kept this scoped to active checks and documented that. Happy to look at a passive variant separately if you think it's worth it (it'd likely need hooking into `countFailure` to detect the threshold crossing).

3. **Naming** — agreed, concise is better. Renamed to `close_if_unhealthy` (both the Caddyfile directive and the JSON field). Docs, adapt test and unit tests updated accordingly.

vnxme:
Thanks. Re point 2, I'm OK with keeping it in `ActiveHealthChecks` unless/until somebody requests anything similar for `PassiveHealthChecks` and explains the underlying behaviour one would like the `proxy` handler to implement.

tannevaled:
Sounds good — thanks! I'll leave it scoped to active checks, and we can revisit a passive variant if a concrete use case comes up.

PR #426, opened 2026-06-01
l4proxy: add request headers and body matching to HTTP health checks

tannevaled:
## What

Extends the active HTTP health check (added in #423) with two commonly-needed knobs:

- **Request headers** — `headers` (JSON) / `health_header <name> <value>` (Caddyfile, repeatable). A `Host` entry sets the request's `Host`.
- **Body matching** — `expect_body` (JSON) / `health_expect_body <regexp>` (Caddyfile). A regular expression the response body must match for the upstream to be considered healthy, in addition to `expect_status`.

The body is only read when `expect_body` is set (otherwise a bounded drain, as before); the regexp is compiled once and cached.

## Why

Some health endpoints need a specific `Host`/auth header, or distinguish state via the body rather than the status code. For example, a check can require both `200` **and** a body matching `"role":"master"`.

## Note on ordering

> ⚠️ **Stacked on #423** (active HTTP health checks). Until that merges, the diff here includes its commit; I'll rebase once #423 is in. Happy to hold this until then.

## Tests

`httpheaders_test.go`: headers + Host are sent, body match / mismatch, invalid `expect_body` regexp errors, and Caddyfile parsing (happy + error paths). `go test ./modules/l4proxy/` passes; `gofmt` / `go vet` / `golangci-lint` clean.

### DIFF
--- a/docs/handlers/proxy.md
+++ b/docs/handlers/proxy.md
@@ -56,6 +56,22 @@ Active health check options include `health_interval`, `health_port` and `health
 - `health_rise` is the number of consecutive successful active health checks required to mark an unhealthy upstream
   healthy again (by default, `1`).
 
+By default an active health check is a raw TCP dial. It can instead probe over HTTP, which lets the proxy follow an
+application-level signal (e.g. a primary-election endpoint such as Patroni's `/primary`) rather than mere port
+liveness. The HTTP probe is enabled by setting `health_uri` and is configured with these options, which correspond to
+the `uri`, `expect_status`, `https` and `tls_skip_verify` fields of the `l4proxy.ActiveHealthChecks` structure:
+
+- `health_uri` is the request path for an HTTP `GET` health check (e.g. `/primary`). When set, the check is performed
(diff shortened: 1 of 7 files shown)


vnxme:
Following https://github.com/mholt/caddy-l4/pull/423#issuecomment-4936377034, I can see the following naming issues to fix inconsistencies with the mainline code:
- `health_header` should be `health_headers`
- `health_expect_body` should be `health_body`, while it should stay linked to the `ExpectBody` field

If it's possible to reuse all the mainline code, I doubt there's any need to split the HTTP health checks feature into multiple PRs, i.e. we can do all the required changes in #423. But let's keep this PR until we agree upon the basics.

P.S. It's also possible to keep `health_expect_body` if the corresponding change is accepted in the mainline repository.

tannevaled:
Agreed on the naming — `health_header` → `health_headers`, and `health_expect_body` → `health_body` (still backed by the `ExpectBody` field, matching mainline's `expect_body` JSON key).

And yes: if we align field-for-field with mainline as discussed in #423, there's no reason to keep the HTTP health-check work split. I'll fold the headers/body support into #423 and close this one once we've settled the basics there.

PR #427, opened 2026-06-02
l4proxy: add rise/fall thresholds to active health checks

tannevaled:
## What

Adds HAProxy-style `rise`/`fall` thresholds to the active health checker:

- `fall` (Caddyfile `health_fall <int>`) — number of **consecutive failed** checks before a peer is marked unhealthy.
- `rise` (Caddyfile `health_rise <int>`) — number of **consecutive successful** checks before an unhealthy peer is marked healthy again.

Both default to `1`, which is exactly the current behavior (flip on the first result), so this is fully backward compatible.

## Why

Today the checker calls `setHealthy` on every individual check, so a single transient failure (or success) immediately flips a peer's state. That makes health flap on a brief blip — undesirable, e.g. for database failover where a momentary timeout shouldn't trigger a switchover. `fall 3` / `rise 2` smooths that out.

## Implementation

Per-peer consecutive-success / consecutive-failure streaks tracked under a mutex (`recordActiveCheck`), capped at the threshold so the counters stay bounded. A streak reaching `fall` marks unhealthy; reaching `rise` marks healthy.

## Tests

`risefall_test.go`: streak logic (incl. reset on opposite result and the default-to-1 behavior), an end-to-end fall threshold via `doActiveHealthCheck`, and Caddyfile parsing (happy + duplicate/invalid errors). `go test ./modules/l4proxy/` passes; `gofmt` / `go vet` / `golangci-lint` clean.

### DIFF
--- a/docs/handlers/proxy.md
+++ b/docs/handlers/proxy.md
@@ -38,7 +38,13 @@ Active health check options include `health_interval`, `health_port` and `health
 - `health_port` is the port to use (if different from the upstream's dial address) for active health checks;
 
 - `health_timeout` sets how long to wait for a connection to be established with a peer (one of the dial addresses)
-  before considering it unhealthy (by default, it equals `5s`).
+  before considering it unhealthy (by default, it equals `5s`);
+
+- `health_fall` is the number of consecutive failed active health checks required to mark an upstream unhealthy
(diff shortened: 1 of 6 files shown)


vnxme:
This is a nice addition, and the code looks good to me -- thank you, @tannevaled!

PR #428, opened 2026-06-02
l4proxy: add weighted load balancing

tannevaled:
## What

Adds weighted load balancing to the layer4 proxy:

- **`Upstream.Weight`** (Caddyfile: `weight <int>` inside an `upstream` block). A value `<= 0` is treated as `1`.
- **`weighted_round_robin`** selection policy — smooth weighted round-robin (the same algorithm nginx uses), skipping unavailable peers.

Existing selection policies and configurations are unaffected.

Example:

```
proxy {
	lb_policy weighted_round_robin
	upstream 10.0.0.1:5432 { weight 3 }
	upstream 10.0.0.2:5432 { weight 1 }
}
```

## Why

Backends rarely have identical capacity. Weighted round-robin lets a bigger node take proportionally more traffic — one of HAProxy's most-used features, and something layer4's existing policies (random / round_robin / least_conn / first / ip_hash) couldn't express.

## Tests

`weighted_test.go`: exact proportional distribution over a full WRR period, default-to-1 behavior, skipping unavailable upstreams, all-down returns nil, and Caddyfile parsing (policy + `weight`, happy + error paths). `go test ./modules/l4proxy/` passes; `gofmt` / `go vet` / `golangci-lint` clean.

### DIFF
--- a/docs/handlers/proxy.md
+++ b/docs/handlers/proxy.md
@@ -72,7 +72,9 @@ Load balancing options include `lb_policy`, `lb_try_duration` and `lb_try_interv
   - `random_choose` is a policy that selects two or more available hosts at random, then chooses the one with
     the least load (Caddyfile syntax is `random_choose [<int>]` with the argument setting the count of available
     hosts to be chosen at random before considering their load);
-  - `round_robin` is a policy that selects an upstream based on round-robin ordering.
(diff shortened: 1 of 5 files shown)


vnxme:
It looks good to me as well -- thank you, @tannevaled!

PR #431, opened 2026-06-02
l4proxy: add Prometheus metrics

tannevaled:
## What

Adds Prometheus metrics to the layer4 proxy, which previously exposed none:

- `caddy_layer4_proxy_connections_total{upstream}` — counter of proxied connections
- `caddy_layer4_proxy_active_connections{upstream}` — gauge of in-flight proxied connections
- `caddy_layer4_proxy_upstream_healthy{upstream}` — gauge (1/0) reflecting active health-check state

## Why

Observability is the one big thing the L4 proxy lacked versus more mature TCP load balancers — no way to see connection counts or upstream health from metrics. These three cover the basics (load + health) per upstream.

## Notes (re: de-duplication)

This deliberately does **not** copy the HTTP server's metrics code — it uses the same *mechanism* (`ctx.GetMetricsRegistry()` + `promauto.With(registry)`), so collectors live on the per-instance registry, reset cleanly across reloads, and surface on the existing admin `/metrics` endpoint. Updates are nil-safe (a never-provisioned handler doesn't panic). Labels are kept to a single `upstream` to bound cardinality; happy to adjust names/labels to your conventions.

## Tests

`metrics_test.go`: counter/gauge increments and the open/close lifecycle, the health gauge, nil-safety, and that an active health check sets the health gauge to 0 (dead peer) and 1 (live listener). `go test ./modules/l4proxy/` passes; `gofmt` / `go vet` / `golangci-lint` clean. go.mod change is just promoting `prometheus/client_golang` from indirect to direct.

### DIFF
--- a/docs/handlers/proxy.md
+++ b/docs/handlers/proxy.md
@@ -9,6 +9,16 @@ title: Proxy Handler
 The Proxy handler implements a layer 4 proxy capable of multiple upstreams with load balancing and health checks.
 This handler is at the core of the package functionality and supports both TCP and UDP.
 
+## Metrics
+
+The handler exposes Prometheus metrics on the instance metrics registry (served by Caddy's admin `/metrics`
+endpoint), labeled by `upstream`:
+
+- `caddy_layer4_proxy_connections_total` — counter of connections proxied to an upstream;
+- `caddy_layer4_proxy_active_connections` — gauge of connections currently being proxied to an upstream;
+- `caddy_layer4_proxy_upstream_healthy` — gauge that is `1` when an upstream is healthy and `0` when it is down,
+  as determined by active health checks.
+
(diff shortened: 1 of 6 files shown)


mholt:
Thank you for all your contributions lately. I imagine you're using an LLM to assist?

We will look at them soon!

tannevaled:
Hi @mholt — thanks for the kind note! Yes, co-authored with Claude Opus 4.8. I design, test, and review every change myself before submitting. These PRs implement features I'm missing downstream, so the work is purposeful and scoped. Happy to slow down if the review load is a concern — no rush on your side!

mholt (approved):
Thanks. I don't know much about metrics stuff, but happy to try this!

PR #432, opened 2026-06-02
l4postgres: add a Postgres TLS handler (terminate inbound + re-originate outbound)

tannevaled:
## What

Adds PostgreSQL STARTTLS support to `l4postgres`, in both directions:

- **Inbound — `postgres_starttls` handler**: reads the 8-byte `SSLRequest`, replies `S`, and hands off. Put a `tls` handler next to terminate TLS; the handlers/matchers that follow then see the **cleartext** startup message (`user`, `database`, `application_name`). It reuses the existing `tls` handler (no TLS code duplicated).
- **Outbound — `StartTLSClient(conn, *tls.Config)`**: re-originates TLS to a Postgres upstream that expects `sslmode != disable` (sends `SSLRequest`, expects `S`, performs a `tls.Client` handshake).

## Why

PostgreSQL doesn't begin TLS with a ClientHello — a libpq client sends `SSLRequest` and waits for `S`/`N` before the handshake (see #187). So the generic `tls` handler can't terminate a *classic* Postgres connection on its own. (PostgreSQL 17+ can negotiate TLS directly via ALPN `postgresql`, which the existing `tls` matcher/handler already covers — see `docs/examples/postgres-over-tls.md`; this handler is for the classic `SSLRequest` flow.)

Working example (verified by a `caddyfile_adapt` integration test in this PR):

```caddyfile
{
    layer4 {
        :5432 {
            @pg postgres
            route @pg {
                postgres_starttls
                tls
                proxy upstream.local:5432
            }
        }
    }
}
```

Once cleartext, this composes with content matchers such as those proposed in #188 — e.g. adding a `postgres { database acme }` matcher (from that PR) inside the route after `tls` to route by database. (#188 isn't merged, so it's intentionally left out of the example above.)

## Notes / design

- The **outbound** part is a self-contained function rather than wired into `l4proxy`, to avoid coupling the generic proxy to Postgres. It would slot in behind an opt-in upstream STARTTLS hook — happy to shape that in review, or split the outbound piece into its own PR. **Terminate-only** deployments (backend on a trusted/encrypted network) don't need the outbound side.

## Tests / docs

- `starttls_test.go`: inbound (replies `S` / rejects non-SSLRequest / read & write error branches), outbound (real TLS handshake against a self-signed test server / `N` refusal / write, reply-read, unexpected-reply, and handshake error branches), and Caddyfile parsing. **100% statement coverage** of the new code.
- `integration/caddyfile_adapt/gd_handler_postgres_starttls.caddytest`: verifies the example above adapts to the expected JSON.
- `docs/handlers/postgres_starttls.md`.
- `go test ./... -race`, `gofmt`, `go vet`, `golangci-lint` clean; no go.mod change.

Builds on #187 and complements #188.


vnxme:
Thank you for your latest contributions! Please add [integration tests](https://github.com/mholt/caddy-l4/tree/master/integration/caddyfile_adapt) and [documentation](https://github.com/mholt/caddy-l4/tree/master/docs/handlers). The [Postgres-over-TLS example](https://github.com/mholt/caddy-l4/blob/master/docs/examples/postgres-over-tls.md) may also require updating.

The config example in the why section above seems a bit strange - could you please explain why you need two `postgres` handlers? Neither do they exist, nor are they introduced by the PR -- or am I missing anything? Have you tested this code/example yourself?

tannevaled:
You're right, and thanks for the careful read — the example in the description was wrong as written: it mixed the `postgres` matcher with handlers and used a `postgres { database ... }` matcher form from #188 that isn't merged. I'd sketched it to illustrate the end state rather than adapting it, which is exactly the wrong shortcut. Apologies.

I've fixed it and added what you asked for:

- A correct, #188-independent example (matcher in `@pg postgres`, handlers `postgres_starttls` → `tls` → `proxy` in the route), now **verified by a `caddyfile_adapt` integration test** (`gd_handler_postgres_starttls.caddytest`) — so it's checked to adapt, not asserted by hand.
- Documentation: `docs/handlers/postgres_starttls.md`.
- Test coverage of the new code brought to **100%** (inbound + outbound, including the error branches).

On `docs/examples/postgres-over-tls.md`: it covers the PostgreSQL 17+ **direct-TLS (ALPN `postgresql`)** path, which is a different mechanism from the classic `SSLRequest` flow this handler targets, so I left it as-is and cross-referenced it from the new handler doc instead. Happy to add a short STARTTLS note there too if you'd prefer.

The PR description is updated accordingly. Thanks again — and please tell me if you'd rather I split the outbound `StartTLSClient` piece into its own PR.

modules/l4postgres/starttls.go
```diff
@@ -0,0 +1,144 @@
...
+//
+// Example (Caddyfile):
+//
+//	route {
+//		postgres
+//		postgres_starttls
+//		tls
+//		# downstream matchers now see the cleartext startup message
+//		proxy ...
+//	}
```
vnxme: This example includes a non-existent `postgres` handler.

vnxme (commented):
Thank you for working on it!

vnxme:
Could we also mention this new intermediary handler [here](https://github.com/mholt/caddy-l4/blob/master/docs/handlers.md)?

tannevaled:
Good call — added `postgres_starttls` to the handlers index in `docs/handlers.md` (under intermediary handlers). Thanks for the review!

modules/l4postgres/starttls.go
tannevaled: Good catch — fixed. `postgres` is a matcher, not a handler, so the example now uses it to gate the route (`@pg postgres` → `route @pg { postgres_starttls; tls; proxy }`) instead of listing a bare `postgres` in the handler chain. The docs and the caddyfile_adapt test already used the matcher form; this was only the doc comment in starttls.go.

vnxme:
I've started to doubt if `postgres_starttls` is a right name for the handler. The [protocol flow](https://www.postgresql.org/docs/current/protocol-flow.html) does only mention `SSL`. Could #187 be the source of confusion with regards to `StartTLS`? Or do you have any references mentioning `StartTLS` as a protocol PostgreSQL implements?

Unless there are any references that support `StartTLS`, I would suggest we rename the handler to `postgres_ssl`. We can also consider `postgres_tls` because it's TLS under the hood after all, but the protocol docs use `SSL` for some reason, that's why I would vote for the former option. And we do `SSLRequest` matching with `postgres { ssl enabled }` in #436.

tannevaled:
You're right — I don't have any reference for "StartTLS" as PostgreSQL terminology. The [protocol flow](https://www.postgresql.org/docs/current/protocol-flow.html) and message formats only ever say `SSL`/`SSLRequest`; "StartTLS" is an SMTP/IMAP/XMPP term that #187 (and my naming) borrowed by analogy. Renamed to `postgres_ssl`, which also pairs cleanly with `postgres { ssl enabled }` from #436:

- handler module id `layer4.handlers.postgres_ssl` (Caddyfile directive `postgres_ssl`);
- outbound helper `StartTLSClient` → `SSLClient`;
- `starttls.go`/`_test.go` → `ssl.go`/`_test.go`, docs → "Postgres SSL Handler", handlers index + `caddyfile_adapt` test updated;
- PR title updated too.

No behavior change; `ssl.go` stays at 100% coverage. I also updated the matching example in #436 to use `postgres_ssl`.

francislavoie:
"SSL" has been dead since 1999, TLS is the replacement. I don't like using SSL here even if postgres documentation mistakenly uses that terminology.

vnxme:
@francislavoie, you're right. Then would you second `postgres_tls`?

francislavoie:
Yeah I think so. 

https://www.postgresql.org/docs/current/ssl-tcp.html

> The terms SSL and TLS are often used interchangeably to mean a secure encrypted connection using a TLS protocol. SSL protocols are the precursors to TLS protocols, and the term SSL is still used for encrypted connections even though SSL protocols are no longer supported. SSL is used interchangeably with TLS in PostgreSQL.

tannevaled:
Fair point @francislavoie — TLS it is. Renamed `postgres_ssl` → **`postgres_tls`** (module id `layer4.handlers.postgres_tls`, Caddyfile `postgres_tls`, `TLSClient` helper, `tls.go`, docs "Postgres TLS Handler", index + adapt test, PR title). I kept the word `SSLRequest` only where the prose names the actual wire message, since that's its literal name in the protocol.

For consistency: the #436 matcher currently exposes this as `postgres { ssl enabled }`. Want me to rename that option to `tls` as well (`postgres { tls enabled }`) so the matcher and handler agree? I left it as-is for now since @vnxme spelled out the `ssl <enabled|disabled|*>` syntax explicitly — happy to switch it if you'd both prefer `tls` there too.

vnxme:
@tannevaled, thank you for a prompt reaction. I believe it's a good idea to make #436 consistent as well.

tannevaled:
Done — renamed the #436 matcher option to `tls` as well (`postgres { tls enabled }`), so matcher and handler now agree. Thanks both!

PR #435, opened 2026-06-03
test(l4proxy): add unit tests for Handle (the proxy main loop)

tannevaled:
## What

`Handler.Handle` — the core proxy loop in `modules/l4proxy` — had no direct unit tests; its behavior was only exercised indirectly via integration tests. This PR adds focused unit tests that bring `Handle` to **100% statement coverage**.

## Tests added

Using the existing `net.Pipe` downstream pattern (already used by the `dialPeers` tests) plus a real loopback echo upstream:

- **No upstream available, no retry** → returns `no upstreams available`.
- **Selected upstream unreachable, no retry** → returns the dial error.
- **No upstream available, retry enabled** → loops through `tryAgain`, then returns once `TryDuration` elapses.
- **Selected upstream unreachable, retry enabled** → exercises the dial-failure retry/`continue` branch, then returns the error.
- **Success path** → selects an upstream, dials it, proxies bytes in both directions, and returns `nil` after the downstream connection closes.

A small `newHandleTestConn` helper pairs an in-memory client end with a `layer4.Connection` wrapping the server end.

## Verification

```
go test ./modules/l4proxy/ -race -run TestHandle   # all pass
```

`Handle` goes from 0% → 100% statement coverage. No production code changes.

### DIFF
--- a/modules/l4proxy/proxy_test.go
+++ b/modules/l4proxy/proxy_test.go
@@ -3,6 +3,7 @@ package l4proxy
 import (
 	"context"
 	"fmt"
+	"io"
 	"net"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


vnxme:
This is a reasonable idea to add tests for the core proxy loop. Looks good to me, thanks!

PR #436, opened 2026-06-05
l4postgres: consolidate into one postgres matcher (user/database, client, TLS) — supersedes #188

tannevaled:
## What

This supersedes #188 by @metafeather, **rebased onto the current hardened `MatchPostgres`** on `master`. The original PR added three Postgres StartupMessage matchers, but it predated the maintainers' rewrite of the connection detector, so it no longer applied cleanly. This brings the same matching capabilities onto the current detector and fills the gaps the original left. @metafeather is preserved as `Co-authored-by:` on the commit.

As discussed on #188, this keeps #432 (the `postgres_starttls` handler) as a separate PR on top. Once this lands I'll close #188.

## Matchers

- **`postgres`** gains an optional `user` map (user → allowed databases, `*` as the wildcard user), applied to the StartupMessage parameters.
- **`postgres_client`** matches on the `application_name` parameter.
- **`postgres_ssl`** matches connections that request SSL, or — with `disabled` — that do not.

```caddyfile
postgres {
    user alice planets_db stars_db
    user * public_db
}
postgres_client psql TablePlus
postgres_ssl            # require SSLRequest
postgres_ssl disabled   # require no SSLRequest
```

## How it differs from #188

Faithful to the original design — **module IDs and JSON config shapes are unchanged, so existing configs keep working** — but elevated to current standards:

- **Reuses `master`'s DoS-hardened length/payload bounds checks** for all three matchers (via a shared `readFirstMessage` helper) instead of the unchecked `Uint32(head)-4` read, which could underflow on a short message.
- **Adds `json` struct tags** so the documented lowercase keys (`user`, `client`, `disabled`) are actually produced.
- **Adds `UnmarshalCaddyfile`** to all three matchers (the original was JSON-only), with `caddyfile_adapt` integration tests.
- **100% statement coverage** on the package, verified under `-race`. The maintainers' existing 388-line `TestMatchPostgres` is kept and still passes unchanged (behavior of the bare matcher is identical).
- **Documentation** for all three matchers under `docs/matchers/`.

## Verification

```
go test ./modules/l4postgres/ ./integration/ -race   # all pass, 100% coverage of l4postgres
```


vnxme:
Thank you for rearranging, expanding and testing the `postgres` matcher PR we've been postponing for 2 years. 

Following [this comment](https://github.com/mholt/caddy-l4/pull/188#pullrequestreview-2096920293), could we have a single matcher instead of 3 ones? I imagine it could have the following syntax:
```caddyfile
postgres {
    # optional directives to match users and database names
    # user <name> [<database>...]
    user alice planets_db stars_db
    user * public_db

    # optional directive to match the application name
    # client <name> [<name>...]
    client psql TablePlus

    # optional directive to match SSL
    # ssl <enabled|disabled|*>
    # `enabled` requires presence of `SSLReuqest`
    # `disabled` requires absence of `SSLRequest`
    # `*` makes the matcher indifferent to `SSLRequest` (default)
    ssl enabled
}
```

tannevaled:
Done — consolidated into a single `postgres` matcher, exactly the shape you sketched:

```caddyfile
postgres {
    user alice planets_db stars_db
    user * public_db
    client psql TablePlus
    ssl <enabled|disabled|*>
}
```

- `MatchPostgres` now carries `user`, `client` and `ssl`; when more than one is set they're AND-ed.
- `ssl` is the tri-state you described: `enabled` requires an `SSLRequest`, `disabled` requires its absence, and `*` (or an omitted JSON value) is indifferent. `Provision` validates it.
- Dropped `MatchPostgresClient` / `MatchPostgresSSL` and their docs/adapt tests; added a combined adapt test and updated `docs/matchers`.
- Still 100% coverage, verified under `-race`.

One inherent wrinkle worth noting: `user`/`client` live on the StartupMessage while `ssl enabled` matches the `SSLRequest` — they're different messages, so combining `ssl enabled` with a `user`/`client` filter never matches. The matcher just AND-s whatever is configured; I documented that in the matcher doc. @metafeather kept as co-author.

modules/l4postgres/matcher.go
```diff
@@ -149,18 +282,96 @@ func validateStartupMessageFormat(data []byte) bool {
...
+	return params
 }
 
-// UnmarshalCaddyfile sets up the matcher from Caddyfile tokens.
+// UnmarshalCaddyfile sets up the matcher from Caddyfile tokens. Syntax:
+//
+//	postgres
+//
+//	postgres {
+//		# match user/database pairs; repeat for each entry, "*" is the wildcard user
```
vnxme: This is where the linter actually fails. I suggest we drop the last word in line 319.
```suggestion
//		# match user/database pairs; repeat for each entry, "*" is the wildcard
```

docs/matchers/postgres.md
```diff
@@ -12,14 +12,47 @@ It does independent raw packet parsing under the hood.
...
+  The special key `*` applies to any user that is not listed explicitly. An empty
+  database list allows any database for that user.
+- `client` is a list of accepted `application_name` values.
+- `ssl` constrains whether the connection must begin with an `SSLRequest`:
+  `enabled` requires one, `disabled` requires its absence, and `*` (the default)
+  is indifferent.
+
+`user` and `client` are carried only by the StartupMessage, so an `SSLRequest`
+never matches when either is set (and, conversely, `ssl enabled` combined with a
+`user`/`client` filter can never match, since those live on different messages).
```
vnxme, line 29: Just to clarify, do I understand correctly that we can do as follows:
```caddyfile
@pg_ssl {
    postgres {
        ssl enabled
    }
}
route @pg_ssl {
    postgres_starttls
    tls
    subroute {
        @pg_allowed {
            postgres {
                user alice planets_db stars_db
                user * public_db
                client psql TablePlus
            }
        }
        route @pg_allowed {
            proxy upstream.local:5432
        }
    }
}
```

If yes, it's definitely worth mentioning. I would describe it in some readme file in the `docs/examples` directory, and put a reference into the main `README.md`.

We can adjust and extend [this example](https://github.com/mholt/caddy-l4/blob/master/docs/examples/postgres-over-tls.md) or add a cross-reference to a separate readme file describing the feature above, so that we avoid any confusion of PostgreSQL users. I know you mentioned the difference [here](https://github.com/mholt/caddy-l4/pull/432/changes#diff-407d261039f9627e6b9e3f17b05e453b76b5403053cc498849a67c70c0731a3f), but I don't mind if we repeat something in the docs to make things clear.

modules/l4postgres/matcher.go
tannevaled: Applied — dropped the trailing `user` (it paired with the `user` on the next line and tripped dupword). Scanned the rest of the file for the same cross-line pattern; this was the only one.

docs/matchers/postgres.md
tannevaled: Yes, your understanding is exactly right. After `postgres_starttls` + `tls` terminate the connection, the `subroute` sees the decrypted StartupMessage, so the inner `postgres { user/database/client }` matcher filters on it just like on a plaintext connection. `ssl enabled` only ever matches the `SSLRequest` (which carries no parameters), which is why the two are split across the outer route and the subroute.

I documented it as you suggested: added `docs/examples/postgres-matching.md` (plaintext matching + the TLS-terminated `subroute` variant, with a note on the SSLRequest-vs-StartupMessage distinction and a cross-reference to the Postgres-over-TLS example), and linked it from `README.md`. It references the `postgres_starttls` handler from #432, so the two land together.

tannevaled:
Done — renamed the matcher's `ssl` option to `tls` so it lines up with the `postgres_tls` handler:

```caddyfile
postgres {
    user alice planets_db stars_db
    user * public_db
    client psql TablePlus
    tls <enabled|disabled|*>
}
```

JSON key is now `"tls"`. Internal protocol identifiers keep their real names (the `SSLRequest` message, `sslRequestCode`), since those are the literal wire names. Docs, the combined `caddyfile_adapt` test, unit tests and the matching example are all updated; still 100% coverage under `-race`.

vnxme:
Thanks again for working on it!

Issue #372, opened 2026-01-21
Metrics support


AbdelrahmanElawady:
Are there any plans to add metrics support similar to what's in caddy http app (request size, response size, ip, etc...)? It seems some of the data is already here: https://github.com/mholt/caddy-l4/blob/93f52b6a03bac66a4321dd1c5287820e3c2a832c/layer4/server.go#L199-L204 they are also not public so using them in a module to handle metrics is not currently feasible. 

Thanks.

vnxme:
@tannevaled, following #431, would you like to help with adding connection stats metrics?

tannevaled:
Happy to — opened #444.

It does two things: exposes `Connection.BytesRead()` / `BytesWritten()` (the counters existed but weren't accessible, which was the blocker noted here), and adds a passthrough `metrics` handler that records per-connection `connections_total` / `received_bytes_total` / `sent_bytes_total` to Prometheus, using the same instance-registry pattern as #431.

Because layer4 keeps byte accounting on the underlying connection, placing `metrics` early in a route captures the raw on-the-wire traffic even through a later `tls` handler. Kept it to counters for a first cut; glad to switch to histograms or add labels if you'd prefer. @AbdelrahmanElawady this should give you the accessor you asked about.

PR #444, opened 2026-07-03
l4metrics: add a connection metrics handler (#372)

tannevaled:
Addresses #372 (connection stats metrics), following @vnxme's suggestion after #431.

## What

Two small pieces:

1. **Expose byte counters** — `layer4.Connection` already tracks `bytesRead`/`bytesWritten` (they're logged), but they weren't accessible to modules (the blocker noted in #372). This adds `Connection.BytesRead()` / `BytesWritten()` accessors.

2. **A `metrics` handler** (`layer4.handlers.metrics`) — a passthrough handler that counts each connection and, once the rest of the chain finishes, records its bytes received/sent to Prometheus:

   - `caddy_layer4_connections_total`
   - `caddy_layer4_received_bytes_total`
   - `caddy_layer4_sent_bytes_total`

   ```caddyfile
   route {
       metrics
       proxy localhost:5433
   }
   ```

## Notes

- Follows the same instance-registry pattern as the proxy metrics from #431 (`ctx.GetMetricsRegistry()`, reset across reloads, nil-safe).
- Place it early in a route so it wraps the whole chain. Because layer4 keeps byte accounting on the underlying connection, the counts reflect **raw on-the-wire** traffic even when a later `tls` handler terminates the connection.
- The accessors are documented as safe to read once the connection has finished (e.g. after `next.Handle` returns), which is exactly when the handler reads them.

## Verification

```
go test ./modules/l4metrics/ ./layer4/ ./integration/ -race
```

100% statement coverage of the new package; docs (`docs/handlers/metrics.md` + index) and a `caddyfile_adapt` test included. Happy to adjust metric names/shape (e.g. histograms instead of counters) to taste.

### DIFF
--- a/docs/handlers.md
+++ b/docs/handlers.md
@@ -35,7 +35,8 @@ Handlers are categorized into different types based on their relationship with n
 |                       | [**echo**](/docs/handlers/echo.md)                     | Echo server, i.e. sends back exactly what it receives                                           |
 |                       | [**proxy**](/docs/handlers/proxy.md)                   | Layer 4 proxy, capable of multiple upstreams (with load balancing and health checks)            |
 |                       | [**socks5**](/docs/handlers/socks5.md)                 | [SOCKSv5](https://www.rfc-editor.org/rfc/rfc1928) server                                        |
-| Intermediary handlers | [**postgres_tls**](/docs/handlers/postgres_tls.md)     | Negotiating the Postgres `SSLRequest` preamble so `tls` can terminate classic Postgres TLS |
+| Intermediary handlers | [**metrics**](/docs/handlers/metrics.md)               | Recording per-connection traffic metrics (connection count, bytes received/sent) to Prometheus  |
+|                       | [**postgres_tls**](/docs/handlers/postgres_tls.md)     | Negotiating the Postgres `SSLRequest` preamble so `tls` can terminate classic Postgres TLS |
 |                       | [**proxy_protocol**](/docs/handlers/proxy_protocol.md) | Receiving [HAProxy Proxy Protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) |
 |                       | [**tls**](/docs/handlers/tls.md)                       | TLS termination                                                                                 |
 |                       | [**throttle**](/docs/handlers/throttle.md)             | Connection throttling to simulate slowness and latency                                          |
--- a/docs/handlers/metrics.md
+++ b/docs/handlers/metrics.md
@@ -1,0 +1,63 @@
+---
+title: Metrics Handler
+---
+
+# Metrics Handler
+
+## Summary
+
+The Metrics handler is a passthrough handler that records per-connection traffic
+metrics to Prometheus. It counts every connection that passes through it and,
+once the rest of the handler chain has finished, adds the connection's bytes
+received and sent to counters.
+
+Place it early in a route so it wraps the whole chain. Because layer4's byte
+accounting is kept on the underlying connection, the counts reflect the raw
+on-the-wire traffic even when a later handler (such as `tls`) terminates the
+connection.
+
+## Exported metrics
+
+All metrics are registered on Caddy's instance metrics registry (reset across
+config reloads) under the `caddy_layer4` subsystem:
+
+- `caddy_layer4_connections_total` — total number of connections handled through
+  the handler;
+- `caddy_layer4_received_bytes_total` — total number of bytes received from
+  clients;
+- `caddy_layer4_sent_bytes_total` — total number of bytes sent to clients.
+
+## Syntax
+
+The handler has no configurable fields.
+
+### Caddyfile
+
+The handler supports the following syntax:
+```caddyfile
+metrics
+```
+
+An example config that collects connection metrics for TCP port 5432 proxied to
+a Postgres backend:
+```caddyfile
+{
+    layer4 {
+        :5432 {
+            route {
+                metrics
+                proxy localhost:5433
+            }
+        }
+    }
+}
+```
+
+### JSON
+
+JSON equivalent to the handler in the config above:
+```json
+{
+    "handler": "metrics"
+}
+```
--- a/docs/handlers/metrics.md
+++ b/docs/handlers/metrics.md
@@ -61,3 +61,69 @@ JSON equivalent to the handler in the config above:
     "handler": "metrics"
 }
 ```
+
+## Scraping the metrics
+
+The `metrics` handler only *records* into Caddy's instance metrics registry — the
+same registry Caddy uses for its own metrics. Exposing that registry for a
+Prometheus scraper is done with Caddy's built-in metrics endpoints, so nothing
+extra is needed in the `layer4` app itself.
+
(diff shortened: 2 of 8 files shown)


vnxme:
@tannevaled, thank you very much! I didn't expect you could do it a few hours after I tagged you in the feature request.

Suppose I have configured Caddy with the Layer 4 app included to collect these metrics by means of placing the `metrics` handler first in in the route chain. How can I export/use these metrics wherever I'd like to? Do I need anything mentioned [here](https://caddyserver.com/docs/metrics)? I mean it would be great if we could provide a complete example of how to collect, export and use layer4 metrics.

tannevaled:
Thanks! And good question — the short answer is that the `metrics` handler only *records* into Caddy's shared instance registry, so exporting is handled by Caddy's normal metrics endpoints, nothing extra in the `layer4` app.

Two ways, both now documented in `docs/handlers/metrics.md`:

1. **Zero-config:** Caddy permanently mounts `/metrics` on its admin API (`http://localhost:2019/metrics` by default), so `curl -s localhost:2019/metrics | grep caddy_layer4` shows the counters right away.
2. **Dedicated endpoint:** serve the same registry on a normal port with the HTTP app's `metrics` handler (e.g. `:9180 { metrics /metrics }`) and point Prometheus at it.

I added a "Scraping the metrics" section with sample `curl` output, the Caddyfile for the dedicated endpoint, and a Prometheus `scrape_config`. Since it's the one shared registry, the same endpoint also exposes the `proxy` handler's `caddy_layer4_proxy_*` metrics (#431) alongside these `caddy_layer4_*` ones.

tannevaled:
Heads-up: applied the same fix as #446 here — the `metrics` handler shared the same duplicate-registration bug (two handlers → panic). It now registers with `reg.Register` and reuses the existing collectors on `AlreadyRegisteredError`, with a regression test. Still 100% coverage.

vnxme:
Thank you for explaining the essentials of metrics collection both here above and in the code as well as fixing the duplicate registration bug. Overall, the code looks good to me. But I have a few more questions to discuss.

1. Just to reconfirm, will the following multi-server and multi-handler example work fine?

```caddyfile
{
    layer4 {
        tcp/:1234 {
            @http http
            route @http {
                subroute {
                    @localhost http host localhost
                    route @localhost {
                        # server 1, handler 1
                        metrics
                        proxy tcp/localhost:80
                    }
                    route {
                        # server 1, handler 2
                        metrics
                        proxy tcp/backend:80
                    }
                }
            }
            route {
                # server 1, handler 3
                metrics
                echo
            }
        }
        udp/:5678 {
            route {
                # server 2, handler 4
                metrics
                proxy udp/backend:53
            }
        }
    }
    servers {
        listener_wrappers {
            layer4 {
                @tls-custom tls alpn custom
                route @tls-custom {
                    tls {
                        connection_policy {
                            alpn custom
                        }
                    }
                    # server 3, handler 5
                    metrics
                    proxy {
                        upstream tcp/backend:443 {
                            tls_insecure_skip_verify
                        }
                    }
                }
            }
            tls
        }
    }
}

localhost {
    tls {
        issuer internal
    }
    respond "OK" 200
}
```

2. The solution suggested in #431 provides for recording metrics automatically for any `proxy` handler, i.e. no extra options are required in the config -- this is quite handy.

    On the other hand, this solution requires a `metrics` handler to be added to the config every time the user wants to count connections passing through the route (as illustrated above) -- this can be tedious and doesn't fully match Caddy philosophy.

    So I imagine we could eliminate a bare form of `metrics` handler completely in favour of automation -- for now, this is for discussion, and I'd like to know what @mholt, @francislavoie and @mohammed90 think about it.

3. If we make every `route` directive collect metrics automatically, it would create at least two issues we have to consider:
    a) Double counting issue. In the example config above, server 1 has three `route` directives with two of them placed inside the remaining one using the `subroute` feature. So I assume we have to make sure `metrics` handlers are only automatically added to the top-level `route` directives to avoid double counting.

    b) No intermediate metrics issue. In the config above, server 3 has a `metrics` handler below a `tls` handler, i.e. it would record decoded bytes after TLS termination. This is a nice feature, at least for debugging purposes which would be gone if we automatically placed a `metrics` handler inside every `route`. So we probably need a solution to either let the user override this automation, or enable dedicated named metrics -- see more details below.

4. When the config has `metrics` enabled for more than one `route`, whether manually as shown above or automatically, the recorded data include all connections and bytes read/written. It is sometimes interesting to see how many connections get passed through the `layer4` app. But the real value of these metrics vanishes.

    In my view, it's really important to let the users record named metrics which is what an expanded version of `metrics` handler should be used for. Suppose, you would like to separately control connections and bytes read/written for TLS connections with SNI `website.local` and from `10.0.0.0/8`. Then you would have to set up a few matchers and put `metrics <tag>` that would expose `caddy_layer4_<tag>_*` metrics. This way there would be no double counting, yet the entire solution would be very versatile.

tannevaled:
Thanks for the thorough review!

**1.** Yes — I adapted that config (dropping the `@localhost host` matcher, which is HTTP-only — there's no `host` matcher in layer4) with the four `metrics` handlers across the two servers plus the subroute, and it adapts and provisions cleanly. Before the duplicate-registration fix the second handler would have panicked; now they all share the same three collectors. That sharing is exactly the behaviour behind your point 4: every `metrics` handler increments the *same* `caddy_layer4_connections_total` / `…_bytes_total`, so the numbers are an aggregate across the whole app, not per-route.

**2/3.** I can see both sides. Automatic per-route collection (like the HTTP app's global `metrics`) is convenient and more "Caddy-like", but the explicit handler has two properties worth keeping that you already spotted: its **position in the chain matters** (a `metrics` placed after `tls` measures decrypted bytes — a genuine feature, 3b), and it lets the user scope what's counted. If we did automate, I agree it would have to be top-level routes only to avoid the double counting in 3a. I'm happy to go whichever way you, @mholt, @francislavoie and @mohammed90 settle on — including dropping the bare handler if the consensus is full automation.

**4.** Strong +1 on named metrics — that's the real fix for the aggregation problem, and it composes with matchers instead of fighting them. I'd be glad to implement `metrics [<tag>]`: an optional tag that scopes the collectors so, in your example, `metrics website` exposes metrics distinct from an untagged one. Two shapes are possible — a `tag` label on the existing `caddy_layer4_*` family, or a distinct `caddy_layer4_<tag>_*` name as you sketched. I lean slightly towards the label (one metric family, easier to aggregate/filter, no metric explosion), but either is easy.

Want me to add the tag to this PR now, or keep this one as the minimal handler and do the tagging (plus whatever we land on for automation) as a follow-up once the direction in 2/3 is settled? Either works for me.

vnxme:
> dropping the @localhost host matcher, which is HTTP-only — there's no host matcher in layer4

Sorry, that's a typo. It should be `@localhost http host localhost`. Fixed in the code example above.

tannevaled:
No worries — with `@localhost http host localhost` your full example (including the `listener_wrappers` + `tls` server and the `localhost` HTTP site) adapts cleanly, and all five `metrics` handlers provision without the duplicate-registration panic. So it works end to end; the only behavioural note stays the aggregation point (4), which the optional tag would address.

PR #452, opened 2026-07-19
l4proxy: add transparent source IP proxying

### LINKED ISSUE #399: Any plans to implement the IP_TRANSPARENT feature similar to Nginx's `proxy_bind`?
Ref.：
1、https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_bind
2、https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_bind_dynamic
3、https://www.kernel.org/doc/html/latest/networking/tproxy.html

floze-the-genius:
## Summary

- add a per-upstream `transparent` option that binds outbound TCP/UDP connections to the effective downstream client IP
- configure Linux sockets with `IP_TRANSPARENT` / `IPV6_TRANSPARENT`, while rejecting unsupported platforms and conflicting source-address settings
- add Caddyfile adaptation, configuration, address-family, and Linux socket-option regression coverage
- document the required upstream route, socket-match diversion, policy routing, capabilities, firewall considerations, and IPv6 equivalent
- update the aligned `golang.org/x/*` module set required by `x/text v0.39.0`, which fixes the newly reported `GO-2026-5970` failure

Closes #399.

## Use case

This is for upstream services that make authorization, rate-limit, audit, or response decisions from the transport peer address but cannot be changed to parse a PROXY protocol header. Examples include legacy or third-party TCP services with source-IP ACLs and UDP services whose behavior depends on the requester's address. Caddy's `remote_ip` matcher can enforce policy at Caddy, but it cannot make the upstream's own logs or source-IP controls observe the client address. PROXY protocol remains preferable whenever the upstream supports it.

## Validation

- `GOTOOLCHAIN=go1.25.12 go test ./...`
- `GOTOOLCHAIN=go1.25.12 go vet ./...`
- `GOTOOLCHAIN=go1.25.12 go mod verify`
- `GOTOOLCHAIN=go1.25.12 go mod tidy -diff`
- `GOTOOLCHAIN=go1.25.12 go run golang.org/x/vuln/cmd/govulncheck@v1.6.0 ./...`
- `GOTOOLCHAIN=go1.25.12 go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest run --timeout 10m`
- `GOOS=linux GOARCH=amd64 go test -c ./modules/l4proxy`
- `GOOS=windows GOARCH=amd64 go test -c ./modules/l4proxy`
- Linux-built transparent tests executed in a Linux container

## AI assistance

AI-assisted coding tools were used to help draft and review parts of the implementation, tests, and documentation. I reviewed the resulting behavior against Go's socket setup and the Linux transparent-proxy documentation, revised the implementation and deployment guidance, and ran the validation above.


vnxme:
Thanks for your contribution! Could you please indicate in the post above whether any AI was used to prepare this PR?

It would also be helpful if you could describe in detail any real-world usage scenarios requiring the `transparent` option. This is a missing bit of the original discussion in #399.

The documentation mentions *policy routing that returns the upstream's replies through the Caddy host* -- would you please elaborate on it? I also doubt if the Caddy host could accept these replies without extra kernel/firewall tweaks.

floze-the-genius:
Thanks for calling this out. I have added the AI-use disclosure to the PR body and expanded both the use-case explanation and deployment documentation.

The case where `transparent` is useful is an upstream that derives its own authorization, rate limiting, audit logs, or response behavior from the transport peer address but cannot parse a PROXY header. `remote_ip` can let Caddy make a policy decision, but it does not make the backend's own source-IP ACLs or logs see the client. PROXY protocol is still the better option when the upstream supports it.

You are also right that binding the non-local address is not sufficient by itself. The updated docs now show both sides of the return path: the upstream (or its gateway) routes the client prefix through Caddy, and the Caddy host uses the transparent socket match plus fwmark/local policy routing to deliver replies to the socket. It also calls out firewall restrictions, the IPv6 equivalent, and the required `CAP_NET_ADMIN` or `CAP_NET_RAW` capability.

The previously red `govulncheck` was unrelated to this feature and came from the repository-wide `golang.org/x/text v0.38.0` call graph after `GO-2026-5970` was published. Commit `67b4237` updates the aligned `x/*` set through `go mod tidy`; `govulncheck v1.6.0` now reports no called vulnerabilities. The full Go 1.25.12 test suite, vet, module verification, tidy diff, and golangci-lint all pass locally.

floze-the-genius:
Hi @vnxme — following up after the update: the PR now includes the AI-use disclosure, concrete backend use cases, and both sides of the Linux return path, along with Linux tests and cross-platform compile checks. The branch is still current with `master`. Is there any remaining concern I should address?

vnxme:
@floze-the-genius, thank you for pinging me. I will review this PR again, when I have time. Other maintainers may also be willing to share their feedback. Don't worry, we're tracking all open PRs.

Issue #451, opened 2026-07-16
Does proxy_protocol handler support UDP?


felixauringer:
The documentation in https://github.com/mholt/caddy-l4/blob/master/docs/handlers/proxy_protocol.md does not clearly state whether UDP is supported. I know that the used [library](https://github.com/pires/go-proxyproto) supports it but I couldn't get it working with caddy-l4. I only found issues regarding UDP support for proxy protocol in the `proxy` handler, but not for incoming connections.

Does the proxy_protocol handler in Caddy-l4 support UDP?

vnxme:
While I don't use proxy protocol and can't verify it myself, I can see no reason why this app shouldn't work for UDP if the underlying library supports it. If you share your config, tests and results, we will probably be able to help you adjust the config or confirm limitations, bugs, etc.

felixauringer:
## Summary

The setup below works for HTTP and HTTPS with normal TLS but not for HTTP/3. I didn't find any documentation that would suggest that proxying UDP with the proxy protocol needs to be configured differently.

```
$ caddy --version
v2.11.4 h1:XKxkMTgNSizEvKG6QHue6cAsFOteU2qA61w2tKkCWi0=
```

(`xcaddy build --with github.com/mholt/caddy-l4` executed yesterday, I believe)

## Caddyfile

General idea:

- Caddy-l4 on ports 10xxx to add proxy protocol header
- Caddy-l4 on ports 9xxx to receive and send proxy protocol
- "HTTP Caddy" on ports 7xxx to serve some website

```
{
	layer4 {
		# No incoming proxy protocol, only outgoing proxy protocol.
		127.0.0.1:10080 {
			route {
				proxy {
					proxy_protocol v2
					upstream 127.0.0.1:9080
				}
			}
		}
		127.0.0.1:10443 {
			route {
				proxy {
					proxy_protocol v2
					upstream 127.0.0.1:9443
				}
			}
		}
		udp/127.0.0.1:10443 {
			route {
				proxy {
					proxy_protocol v2
					upstream udp/127.0.0.1:9443
				}
			}
		}

		# Incoming and outgoing proxy protocol.
		127.0.0.1:9080 {
			route {
				proxy_protocol {
					allow 127.0.0.1/32
					fallback_policy REJECT
				}
				proxy {
					proxy_protocol v2
					upstream 127.0.0.1:7080
				}
			}
		}
		127.0.0.1:9443 {
			route {
				proxy_protocol {
[...]

PR #453, opened 2026-07-21
proxy_protocol: Support PROXY headers on every UDP datagram

vnxme:
## AI Disclosure

This PR was drafted by Claude Opus 4.8 following the discussion in #451.

My brief curl-based tests show it fixes this issue. But I don't use the proxy protocol much, so I can't test it extensively.

## Description

The proxy_protocol handler delegated to go-proxyproto, whose Conn strips exactly one PROXY header per connection (guarded by a sync.Once). That matches a byte stream, but caddy-l4's own packet writer prepends a header to every datagram, so over UDP only the first datagram was unwrapped and every subsequent datagram's header leaked through as payload.

Add a per-datagram reader for packet connections that strips a PROXY header off each datagram, and route UDP conns to it while leaving the stream (TCP/Unix) path on go-proxyproto unchanged.

- layer4: add Connection.IsPacketConn() and WrapWithEmptyBuffer(). The latter is needed because the packet reader reads through cx, so the wrapped Connection must not also serve the already-prefetched datagrams raw from the shared buffer.
- l4proxyprotocol: add packetConn (packet.go) which parses and strips a header per datagram, buffers leftover payload across Reads, exposes the first header's addresses via RemoteAddr/LocalAddr, and honors the require/passthrough policy; branch Handle to it for packet conns.
- Add tests covering per-datagram stripping, passthrough when no header is required, rejection when required but missing, and payloads larger than the read buffer.

### DIFF
--- a/layer4/connection.go
+++ b/layer4/connection.go
@@ -190,6 +190,27 @@ func (cx *Connection) Wrap(conn net.Conn) *Connection {
 	}
 }
 
+// WrapWithEmptyBuffer is like Wrap, but the returned Connection starts with an
+// empty prefetch buffer instead of inheriting cx's. Use it when conn already
+// reads through cx — and therefore already delivers any bytes cx previously
+// prefetched — so that those bytes are not also served directly from the
+// returned Connection's buffer, bypassing conn. This is required by the packet
+// (e.g. UDP) PROXY protocol reader, which must strip a header from every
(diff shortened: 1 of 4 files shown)


mohammed90:
If I may, @WeidiDeng, I remember you doing some work on PROXY protocol and UDP. I don't remember the details. You might have insights.

PR #455, opened 2026-07-30
layer4: Don't abort matching when a later route conclusively fails

Pizmovc:
## Problem

A route that reaches a **definitive non-match from partially delivered data** aborts matching for earlier routes that are still waiting for more bytes. The connection falls through to the next handler in microseconds, long before `matching_timeout`, with no timeout logged.

`RouteList.Compile` is shared by the `layer4` app and the listener wrapper, so both are affected. Reproduces on v0.1.1, v0.1.2 and `master`.

## Root cause

`lastNeedsMoreIdx` is overloaded — it is both "index of a route that needs more data" and a watermark for "how far the list is conclusively decided". The non-match branch advances it unconditionally (`lastNeedsMoreIdx = i + 1`), burying the pending state of an earlier route. When the deciding route is the **last** in the list, the watermark reaches `len(routes)` and the retry guard fails:

```go
// some of the matchers can't reach a conclusion
if lastNeedsMoreIdx != -1 && lastNeedsMoreIdx < len(routes) {
    goto loop
}
```

`routesStatus` still holds `routeNeedsMore` for the earlier route, but it is never consulted — so the loop exits without ever prefetching again.

## Why `tls` + `http` triggers it

`isHttp()` rules a stream out as soon as the prefetched bytes contain a `0x0a` at index >= 10. A binary ClientHello reliably has one early (cipher suite `0xc00a`, extension `0x000a`, or the random bytes) — around offset 149 in an OpenSSL 3 hello.

So with an `http` route after a `tls` route: the `tls` matcher asks for more data, the trailing `http` route concludes "not HTTP" and buries that request, and matching is abandoned. Any ClientHello split across TCP segments is affected. Post-quantum hellos (~1700–2400 B) always exceed a typical MSS, so it becomes a race on whether segments coalesce before the first prefetch — intermittent handshake failures, and with a plaintext fallback on `:443`, an HTTP 400 answering a ClientHello (`SSL_ERROR_RX_RECORD_TOO_LONG` / `WRONG_VERSION_NUMBER`).

## Fix

Consult `routesStatus`, which already tracks pending state accurately, instead of relying on the watermark. Additive — existing watermark semantics and route-skipping are untouched. Termination is unaffected: whenever a route still needs data, `lastNeedsMoreIdx != -1`, so the loop still prefetches, and prefetch either progresses or hits the read deadline.

## Test

`TestHttpMatchingDoesNotAbortTlsMatching` feeds a real ClientHello in two writes 100 ms apart, cut just past the first `0x0a` at index >= 10. Without the fix only the `tls-before-http` case fails; `http-before-tls` and `tls-only` pass either way, isolating the ordering dependency rather than a prefetch-count problem. `go test ./...` passes with the fix.

Also verified end-to-end on FreeBSD amd64 (OPNsense, Caddy v2.11.4) with two binaries differing only in this patch. Fall-through on a non-matching SNI, plain-HTTP route matching, and `matching_timeout` firing at 5.00 s are identical between them.

<details>
<summary>Self-contained reproducer — config, script, and before/after output</summary>

No upstream needed: the matched route uses `echo`, so a matched connection gets its own ClientHello mirrored back (first byte `0x16`) while an abandoned one gets a plaintext HTTP response from the fall-through site.

`Caddyfile`:

```caddyfile
{
	auto_https off
	admin off
	skip_install_trust
	log {
		level ERROR
	}

	servers {
		listener_wrappers {
			layer4 {
				matching_timeout 5s

				@sni tls sni example.test
				route @sni {
					echo
				}

				@h http host example.test
				route @h {
					echo
				}
			}
			tls
		}
	}
}

:8443 {
}
```

`repro.py`:

```python
#!/usr/bin/env python3
"""Reproducer for: a conclusively-unmatched route aborts matching for earlier routes.

Run the Caddyfile above, then this script. It sends a real ClientHello for a
matching SNI, first in one write, then split across two writes with a pause.

  '.' = ClientHello echoed back  -> the tls route matched
  'X' = plaintext HTTP response  -> matching was abandoned, fell through
"""
import socket, ssl, time

HOST, PORT, SNI, ATTEMPTS, GAP = "127.0.0.1", 8443, "example.test", 4, 0.4


def client_hello(sni):
    ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
    ctx.check_hostname = False
    ctx.verify_mode = ssl.CERT_NONE
    inb, outb = ssl.MemoryBIO(), ssl.MemoryBIO()
    obj = ctx.wrap_bio(inb, outb, server_hostname=sni)
    try:
        obj.do_handshake()
    except ssl.SSLWantReadError:
        pass
    return outb.read()


def attempt(hello, cut):
    s = socket.create_connection((HOST, PORT), timeout=10)
    try:
        if cut:
            s.sendall(hello[:cut])
            time.sleep(GAP)
            s.sendall(hello[cut:])
        else:
            s.sendall(hello)
        return "." if s.recv(4096)[:1] == b"\x16" else "X"
    except Exception:
        return "X"
    finally:
        s.close()


hello = client_hello(SNI)
print("ClientHello = %d bytes, SNI %s" % (len(hello), SNI))

for label, cut in [("single write", 0)] + [("split at %dB" % c, c) for c in (50, 100, 200, 512, 1400)]:
    r = "".join(attempt(hello, cut) for _ in range(ATTEMPTS))
    print("  %-16s %s  matched=%d/%d" % (label, r, r.count("."), ATTEMPTS))
```

Before (v0.1.2):

```
ClientHello = 1519 bytes, SNI example.test
  single write     ....  matched=4/4
  split at 50B     ....  matched=4/4
  split at 100B    ....  matched=4/4
  split at 200B    XXXX  matched=0/4
  split at 512B    XXXX  matched=0/4
  split at 1400B   XXXX  matched=0/4
```

After (this patch):

```
ClientHello = 1519 bytes, SNI example.test
  single write     ....  matched=4/4
  split at 50B     ....  matched=4/4
  split at 100B    ....  matched=4/4
  split at 200B    ....  matched=4/4
  split at 512B    ....  matched=4/4
  split at 1400B   ....  matched=4/4
```

The 50/100 B rows pass either way because `isHttp()` has not seen a `0x0a` yet and also asks for more data, which keeps the loop alive. Reordering the two routes so `@h` comes first also makes every row pass without the patch.

</details>

## Workaround

Order `http` routes before `tls` routes. They are mutually exclusive on wire format, so this changes no routing outcome.

## AI Disclosure

I used Claude in the process of drafting this PR.


mkallfass:
Independent confirmation of both the bug and the fix's effectiveness, from a
different setup: OPNsense 26.7 (FreeBSD amd64) with the `caddy-custom` package
/ `os-caddy` plugin, which embeds caddy-l4 as listener wrappers.

**Setup:** six `listener_wrappers` routes on the same listener, mixing
`tls sni` and `http host` matchers (TLS SNI passthrough to internal backends
plus plain-HTTP host routing) — a pattern the OPNsense plugin UI naturally
produces. The last route in the chain was an `http`-matcher route, i.e.
exactly the constellation this PR describes.

**Symptom:** ~80–93 % of inbound TLS connections on the SNI-passthrough
hostnames were closed immediately after the TCP handshake — FIN 5–15 ms after
SYN-ACK, before any TLS data was proxied. Bypassing the proxy: 0 % failures.
`service caddy restart` never helped. Failure rates tracked ClientHello
segmentation as analyzed here: modern clients send post-quantum ClientHellos
(X25519MLKEM768, ~1.7–2.4 kB > MSS, so split across TCP segments); small
single-segment ClientHellos went through fine. Observed rates differed by
path (IPv4 ~80 %, IPv6 ~40 %), consistent with a race on whether both
segments land in the same read.

**Persistence:** reproduced unchanged across four package versions
(Caddy 2.11.2 → 2.11.3 → 2.11.3.1 → 2.11.4) and an OS major upgrade
(OPNsense 26.1 → 26.7, new FreeBSD base) — consistent with a logic bug in
the matching loop rather than anything platform- or version-specific.

30-connection probe (`openssl s_client` against the listener, fixed SNI):

| caddy-l4 host build | Result |
|---|---|
| Caddy 2.11.3 | ok=2 / fail=28 |
| Caddy 2.11.3.1 (pkg respin) | ok=4 / fail=26 |
| Caddy 2.11.4 | ok=4 / fail=26 |
| Caddy 2.11.4, new OS major | ok=5 / fail=25 |

**Workaround validation:** reordering the routes so that all `http`-matcher
routes precede all `tls sni` routes (leaving "need more data" routes at the
tail of the chain) immediately resolved it — ok=30 / fail=0 on the previously
failing SNI, 15/15 on two further SNI routes, and the plain-HTTP routes still
route correctly. A pure reorder would have changed nothing if the root cause
were anywhere else, so this is strong empirical confirmation of the mechanism
fixed by this PR.

Thanks @Pizmovc for tracking this down — this has been silently breaking
real-world SNI passthrough setups behind OPNsense for months, and it will
bite more users as post-quantum ClientHellos become the default. Happy to
test a build or provide packet captures if useful for review.

